PR #35, opened 2020-07-15
ensure file object consistency in firefox

### LINKED ISSUE #290: Directory drag-and-drop on Firefox does not preserve Content-Type
There is an issue on Firefox where drag-and-dropping a directory for upload will lose all files Content-type. Files will be incorrectly uploaded with the content-type `application/octet-stream`   
Tested on Firefox 78.0.1 on MacOS but also reported to happen on Windows from another user on Discord.

### Steps to reproduce
 
1. Put 1 or more images of standard mime type file in a folder (jpg, png...)
2. Drag-and-drop the folder in the Drop zone in a Firefox browser
3. Once uploaded, click on the generated Skylink. The browser will download the image instead of displaying it because of improper mime type

### Things to consider

* The bug only happens on drag-and-drop. Clicking on the zone and selecting the directory will keep the proper mime-type
* The bug happens whether the dropzone is in "File mode" or "Directory mode". But will only happen when drag-and-dropping a directory from the file system.
* This happens because the content type is `application/octet-stream` during the multipart upload (so it's an upload issue)

### Progress

I tried digging down and find the source of the bug, from my findings it might be an issue with Firefox itself, and I could not find any workaround.  
I first dug into the [react-dropzone](https://github.com/react-dropzone/react-dropzone) dependency to understand how it was managing dropped directories, only to find that [it was using a sub-dependency](https://github.com/react-dropzone/react-dropzone/blob/v11.0.1/src/index.js#L361).  
And when digging down into [file-selector](https://github.com/react-dropzone/file-selector) I found out that it converts the [DataTransfer](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/DataTransfer) from the 'drop' event into files using [FileSystemFileEntry.file()](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileEntry/file). Using that method leaves the File.type empty on Firefox.

From there I tried to workaround that limitation a bit or find another way to get the files with mime type from the drop event, with no success. It's all the time I had today to work on this but I wanted to document the issue and my progress here so someone or I can pick it up later.

### Conclusion + possible solution

So right now it mostly seems like a bug with Firefox `FileSystemFileEntry`, other drag-and-drop library might have the exact same problem as it seems like the standard way to manage drag-and-drop uploads right now.

A possible solution would be to use another library like [mime-types](https://github.com/jshttp/mime-types) or [mime-db](https://github.com/jshttp/mime-db) to detect the files mime-type when the `File.type` attribute is empty. But I'm not sure if we can manually change the Content-type in a multipart upload after that.

interstella-5555:
Sometimes file object might have had the type property defined menually with Object.defineProperty and some browsers (namely firefox) can have problems reading it after the file has been appended to form data. To overcome this, we recreate the file object using native File constructor with a type defined as a constructor argument.

Closes https://github.com/NebulousLabs/skynet-webportal/issues/290

### DIFF
--- a/src/upload.js
+++ b/src/upload.js
@@ -4,7 +4,7 @@ import parse from "url-parse";
 export async function upload(portalUrl, file, options = {}) {
   const formData = new FormData();
 
-  formData.append("file", file);
+  formData.append("file", ensureFileObjectConsistency(file));
 
   const parsed = parse(portalUrl);
 
@@ -29,7 +29,7 @@ export async function uploadDirectory(portalUrl, directory, filename, options = 
   const formData = new FormData();
 
   Object.entries(directory).forEach(([path, file]) => {
-    formData.append("files[]", file, path);
+    formData.append("files[]", ensureFileObjectConsistency(file), path);
   });
 
   const parsed = parse(portalUrl);
@@ -50,4 +50,15 @@ export async function uploadDirectory(portalUrl, directory, filename, options = 
   );
 
   return data;
+}
+
+/**
+ * Sometimes file object might have had the type property defined menually with Object.defineProperty
+ * and some browsers (namely firefox) can have problems reading it after the file has been appended
+ * to form data. To overcome this, we recreate the file object using native File constructor with
+ * a type defined as a constructor argument.
+ * Related issue: https://github.com/NebulousLabs/skynet-webportal/issues/290
+ */
+export function ensureFileObjectConsistency(file) {
+  return new File([file], file.name, { type: file.type });
 }
--- a/src/upload.js
+++ b/src/upload.js
@@ -53,12 +53,12 @@ export async function uploadDirectory(portalUrl, directory, filename, options = 
 }
 
 /**
- * Sometimes file object might have had the type property defined menually with Object.defineProperty
+ * Sometimes file object might have had the type property defined manually with Object.defineProperty
  * and some browsers (namely firefox) can have problems reading it after the file has been appended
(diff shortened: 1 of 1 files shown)


src/upload.js
```diff
@@ -51,3 +51,14 @@ export async function uploadDirectory(portalUrl, directory, filename, options =
 
   return data;
 }
+
+/**
+ * Sometimes file object might have had the type property defined menually with Object.defineProperty
```
mrcnski:
```suggestion
 * Sometimes file object might have had the type property defined manually with Object.defineProperty
```

src/upload.js
```diff
@@ -51,3 +51,14 @@ export async function uploadDirectory(portalUrl, directory, filename, options =
...
+ * and some browsers (namely firefox) can have problems reading it after the file has been appended
+ * to form data. To overcome this, we recreate the file object using native File constructor with
+ * a type defined as a constructor argument.
+ * Related issue: https://github.com/NebulousLabs/skynet-webportal/issues/290
+ */
+export function ensureFileObjectConsistency(file) {
```
mrcnski: Is there any need to export the function?

mrcnski (changes requested):

src/upload.js
interstella-5555:
```suggestion
function ensureFileObjectConsistency(file) {
```

mrcnski (approved):

PR #40, opened 2020-07-20
Refactor and add default options

mrcnski:
A minor refactor changing the way URLs are formed internally and adding default
options for upload and download. In addition to bringing the SDK more in line
with the other ones, this refactor makes it easier to implement more endpoint
functions s well as to add encryption to uploads and downloads.


README.md
```diff
@@ -165,7 +165,7 @@ Currently supported string types are:
 - direct skylink string, example "XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg"
 - sia: prefixed string, example "sia:XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg"
 - sia:// prefixed string, example "sia://XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg"
-- skylink from url, example "https://siasky.net/XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg"
+- skylink from url, for example `https://siasky.net/XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg`
```
interstella-5555: nit: any reason why this example was changed and it doesn't conform to the how the other examples are defined?

src/download.js
```diff
@@ -1,7 +1,35 @@
...
+export function getDownloadUrl(portalUrl, skylink, customOptions = {}) {
+  const opts = { ...defaultDownloadOptions, ...customOptions };
+
+  let query = {};
+  if (customOptions.download) {
+    query = { attachment: true };
+  }
+  const url = makeUrlWithSkylink(portalUrl, opts.portalEndpointPath, skylink, query);
+
+  return url;
```
interstella-5555:
```suggestion
  return makeUrlWithSkylink(portalUrl, opts.portalEndpointPath, skylink, query);
```

src/download.js
```diff
@@ -1,7 +1,35 @@
...
-export function download(portalUrl, skylink) {
-  const url = getUrl(portalUrl, parseSkylink(skylink), { download: true });
+export const defaultDownloadOptions = {
+  ...options,
+  portalEndpointPath: "/",
+};
+
+export function download(portalUrl, skylink, customOptions = {}) {
+  const opts = { ...defaultDownloadOptions, ...customOptions };
+  opts.download = true;
```
interstella-5555:
```suggestion
  const opts = { ...defaultDownloadOptions, ...customOptions, download: true };
```

src/download.js
```diff
@@ -1,7 +1,35 @@
...
+  window.open(url, "_blank");
+}
+
+export function getDownloadUrl(portalUrl, skylink, customOptions = {}) {
+  const opts = { ...defaultDownloadOptions, ...customOptions };
+
+  let query = {};
+  if (customOptions.download) {
+    query = { attachment: true };
+  }
```
interstella-5555:
```suggestion
  const query = opts.download ? { attachment: true } : {};
```

src/download.js
```diff
@@ -1,7 +1,35 @@
...
+  const url = makeUrlWithSkylink(portalUrl, opts.portalEndpointPath, skylink, query);
+
+  return url;
+}
+
+export function open(portalUrl, skylink, customOptions = {}) {
+  const opts = { ...defaultDownloadOptions, ...customOptions };
+
+  const url = makeUrlWithSkylink(portalUrl, opts.portalEndpointPath, skylink);
```
interstella-5555:
```suggestion
  const opts = { ...defaultDownloadOptions, ...customOptions };
  const url = makeUrlWithSkylink(portalUrl, opts.portalEndpointPath, skylink);
```

src/utils.js
```diff
@@ -2,24 +2,29 @@ import parse from "url-parse";
 
 export const defaultPortalUrl = "https://siasky.net";
 
-export function open(portalUrl, skylink) {
-  const url = getUrl(portalUrl, parseSkylink(skylink));
+export const options = {
```
interstella-5555:
```suggestion
export const defaultGlobalOptions = {
```

src/utils.js
```diff
@@ -2,24 +2,29 @@ import parse from "url-parse";
...
+  portalEndpointPath: "",
+  // TODO:
+  // customUserAgent: "",
+};
 
-  window.open(url, "_blank");
-}
-
-export function getUrl(portalUrl, skylink, options = {}) {
+export function makeUrl(portalUrl, endpointPath, query = {}) {
```
interstella-5555:
```suggestion
export function makeUrl(portalUrl, pathname, query = {}) {
```

it's not always "endpoint path" in your usages

src/download.js
```diff
@@ -1,7 +1,35 @@
-import { getUrl, parseSkylink } from "./utils.js";
+import { makeUrlWithSkylink, options } from "./utils.js";
 
-export function download(portalUrl, skylink) {
-  const url = getUrl(portalUrl, parseSkylink(skylink), { download: true });
+export const defaultDownloadOptions = {
+  ...options,
+  portalEndpointPath: "/",
```
interstella-5555: I don't think you need that defined here, should work without it

src/upload.js
```diff
@@ -1,50 +1,67 @@
...
 
-  const parsed = parse(portalUrl);
+  var filename;
+  if (opts.customDirname != "") {
+    filename = opts.customDirname;
+  }
+  else {
+    var file; for (file in directory) break;
+    filename = getRootDirectory(directory[file]);
+  }
```
interstella-5555:
```suggestion
  // either use a custom dir name or extract it from one of the files
  const filename = opts.customDirname ?? getRootDirectory(Object.values(directory)[0]);
```

src/utils.js
```diff
@@ -1,25 +1,50 @@
...
 }
 
+export function makeUrlWithSkylink(portalUrl, endpointPath, skylink, query = {}) {
+  skylink = parseSkylink(skylink);
+  // Right-trim any forward slashes from endpoint path.
+  endpointPath = endpointPath.replace(/\/+$/g, "");
+  // Left-trim any forward slashes from skylink.
+  skylink = skylink.replace(/^\/+/g, "");
+  endpointPath = `${endpointPath}/${skylink}`;
+  return makeUrl(portalUrl, endpointPath, query);
```
interstella-5555: I think it could be just this because `path.poisx.join` will take care of any excessive `/` between `endpointPath` and `skylink` and `makeUrl` will properly merge the pathname with the hostname.
```suggestion
  const parsedSkylink = parseSkylink(skylink);
  
  return makeUrl(portalUrl, path.posix.join(endpointPath, parsedSkylink), query);
```

interstella-5555 (changes requested):

README.md
mrcnski: I was getting a markdownlint error about a bare URL . I thought making it a code block would also be more readable. I'll make it consistent with the other examples

src/utils.js
mrcnski: I don't know about `Global` since the options can be different for different function calls, e.g. you can try to upload to two different portals simultaneously. I'll call it `defaultOptions`

src/download.js
mrcnski: I know, but I'd rather be explicit with all the default options to avoid confusion.

interstella-5555 (approved):

PR #48, opened 2020-07-30
Add openHns method

mrcnski:


src/download.js
```diff
@@ -1,35 +1,79 @@
...
   const url = this.getDownloadUrl(skylink, opts);
 
   window.open(url, "_blank");
 };
 
+SkynetClient.prototype.downloadHns = async function (hns, customOptions = {}) {
+  const opts = { ...defaultDownloadHnsOptions, ...customOptions };
+};
+
+SkynetClient.prototype.downloadHnsres = async function (hnsres, customOptions = {}) {
```
interstella-5555:
```suggestion
SkynetClient.prototype.resolveHns = async function (domainName, customOptions = {}) {
```

src/download.js
```diff
@@ -1,35 +1,79 @@
...
+ * @param {string} [customOptions.endpointPath="/"] - The relative URL path of the portal endpoint to contact.
+ */
 SkynetClient.prototype.download = function (skylink, customOptions = {}) {
   const opts = { ...defaultDownloadOptions, ...customOptions, download: true };
   const url = this.getDownloadUrl(skylink, opts);
 
   window.open(url, "_blank");
 };
 
+SkynetClient.prototype.downloadHns = async function (hns, customOptions = {}) {
```
interstella-5555:
```suggestion
SkynetClient.prototype.downloadHns = async function (domainName, customOptions = {}) {
```

src/download.js
```diff
@@ -1,35 +1,79 @@
...
+};
+
 SkynetClient.prototype.getDownloadUrl = function (skylink, customOptions = {}) {
   const opts = { ...defaultDownloadOptions, ...customOptions };
   const query = opts.download ? { attachment: true } : {};
 
-  return makeUrlWithSkylink(this.portalUrl, opts.endpointPath, skylink, query);
+  let url = makeUrl(this.portalUrl, opts.endpointPath, parseSkylink(skylink));
+  url = addUrlQuery(url, query);
+  return url;
```
interstella-5555:
```suggestion
  const url = makeUrl(this.portalUrl, opts.endpointPath, parseSkylink(skylink));
  
  return addUrlQuery(url, query);
```

src/download.test.js
```diff
@@ -11,6 +16,7 @@ const validSkylinkVariations = [
   `${portalUrl}/${skylink}/foo/bar`,
   `${portalUrl}/${skylink}?foo=bar`,
 ];
+const validHnsLinkVariations = [hnsLink, `hns:${hnsLink}`, `hns://${hnsLink}`];
```
interstella-5555, line 22: I don't think those `hns:` and `hns://` are valid...

src/upload.js
```diff
@@ -50,7 +50,8 @@ SkynetClient.prototype.uploadDirectory = async function (directory, filename, cu
     formData.append(opts.portalDirectoryFileFieldname, ensureFileObjectConsistency(file), path);
   });
 
-  const url = makeUrl(this.portalUrl, opts.endpointPath, { filename });
+  let url = makeUrl(this.portalUrl, opts.endpointPath);
+  url = addUrlQuery(url, { filename });
```
interstella-5555: could you add query argument to makeUrl so it's all in one function?

src/utils.js
```diff
@@ -1,9 +1,20 @@
 // import axios from "axios";
 import path from "path-browserify";
 import parse from "url-parse";
+import urljoin from "url-join";
 
 export const defaultSkynetPortalUrl = "https://siasky.net";
 
+export const uriHandshakePrefix = "hns:";
+export const uriHandshakeResolverPrefix = "hnsres:";
```
interstella-5555, line 9: Are those valid? I haven't seen or used those myself

src/utils.js
```diff
@@ -56,27 +67,23 @@ export function getRootDirectory(file) {
...
-  return makeUrl(portalUrl, path.posix.join(endpointPath, parsedSkylink), query);
+/**
+ * Properly joins paths together to create a URL. Takes a variable number of
+ * arguments.
+ */
+export function makeUrl() {
+  let args = Array.from(arguments);
+  return args.reduce(function (acc, cur) {
+    return urljoin(acc, cur);
+  });
```
interstella-5555: I think this function doesn't make sense, it's doing exactly the same as `urljoin`.

`makeUrl('/a/', '/b/', '/c/')` === `urljoin('/a/', '/b/', '/c/')` ?

src/utils.js
```diff
@@ -97,3 +103,16 @@ export function parseSkylink(skylink = "") {
 
   throw new Error(`Could not extract skylink from '${skylink}'`);
 }
+
+export function trimUriPrefix(str, prefix) {
+  const longPrefix = `${prefix}//`;
+  if (str.indexOf(longPrefix) == 0) {
```
interstella-5555:
```suggestion
  if (str.startsWith(longPrefix)) {
```

src/utils.js
```diff
@@ -97,3 +103,16 @@ export function parseSkylink(skylink = "") {
...
+    // longPrefix is exactly at the beginning
+    return str.slice(longPrefix.length);
+  }
+  if (str.indexOf(prefix) == 0) {
```
interstella-5555:
```suggestion
  if (str.startsWith(prefix)) {
```

interstella-5555 (changes requested):

src/utils.js
mrcnski: David gave me these prefixes.

src/upload.js
mrcnski: I don't think so, because `makeUrl` takes a variable number of arguments. Would be nice if it was possible, though.

src/utils.js
mrcnski: `urljoin` failed the unit tests. I think if one of the arguments in the middle is `/` it will append it to a slash after it. I'll document this

src/download.js
mrcnski: Hmm, it doesn't just resolve, it downloads too. Most accurate would be `resolveAndDownloadHns` but that's pretty long.

README.md
```diff
@@ -70,10 +70,9 @@ With authentication:
 ```javascript
 import { SkynetClient } from "skynet-js";
 
-const client = new SkynetClient("https://siasky.net");
-
 async function authenticationExample() {
   try {
+    const client = new SkynetClient("https://siasky.net");
```
mrcnski: I'll revert this, wasn't intended.

package.json
```diff
@@ -9,7 +9,7 @@
     "src/*"
   ],
   "scripts": {
-    "test": "jest",
+    "test": "jest --runInBand",
```
interstella-5555, line 12: any reason to run in band?

src/client.js
```diff
@@ -1,10 +1,57 @@
...
+   * The Skynet Client which can be used to access Skynet.
+   * @constructor
+   * @param {string} [portalUrl="https://siasky.net"] - The portal URL to use to access Skynet, if specified.
+   */
+  constructor(portalUrl = "") {
+    if (portalUrl === "") {
       portalUrl = defaultPortalUrl();
     }
     this.portalUrl = portalUrl;
   }
```
interstella-5555:
```suggestion
  constructor(portalUrl = defaultPortalUrl()) {
    this.portalUrl = portalUrl;
  }
```

src/download.js
```diff
@@ -1,35 +1,83 @@
...
 SkynetClient.prototype.download = function (skylink, customOptions = {}) {
-  const opts = { ...defaultDownloadOptions, ...customOptions, download: true };
+  const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions, download: true };
   const url = this.getDownloadUrl(skylink, opts);
 
   window.open(url, "_blank");
 };
 
+SkynetClient.prototype.downloadHns = async function (domain, customOptions = {}) {
+  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
```
interstella-5555:
```suggestion
  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
  
  throw new Error("Unimplemented");
```

src/download.js
```diff
@@ -1,35 +1,83 @@
...
+};
+
+SkynetClient.prototype.downloadHnsres = async function (domain, customOptions = {}) {
+  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
```
interstella-5555:
```suggestion
  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
  
  throw new Error("Unimplemented");
```

src/download.js
```diff
@@ -1,35 +1,83 @@
...
-SkynetClient.prototype.metadata = function (skylink, customOptions = {}) {
-  const opts = { ...defaultDownloadOptions, ...customOptions };
+SkynetClient.prototype.metadata = async function (skylink, customOptions = {}) {
+  const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
 
   throw new Error("Unimplemented");
 };
 
+/**
+ * Opens the content at the skylink within the browser.
```
interstella-5555:
```suggestion
 * Opens the content of the skylink within the browser.
```

src/download.js
```diff
@@ -1,35 +1,83 @@
...
+  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
+
+  domain = trimUriPrefix(domain, uriHandshakePrefix);
+
+  // Get the skylink from the hns domain on the portal.
+  const response = await this.executeRequest({ ...opts, method: "get", extraPath: domain });
+  const skylink = response.data.skylink;
+
+  this.open(skylink, customOptions);
+};
```
interstella-5555: I would probably consider extending `open` to accept handshake domains with some param like `{ handshake: true }` instead of creating clones of the `open` and `download` functions

src/download.js
```diff
@@ -1,35 +1,83 @@
...
+
+SkynetClient.prototype.openHnsres = async function (domain, customOptions = {}) {
+  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
```
interstella-5555:
```suggestion
  const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
  
  throw new Error("Unimplemented");
```

src/download.test.js
```diff
@@ -1,6 +1,12 @@
+import axios from "axios";
+import MockAdapter from "axios-mock-adapter";
+
 import { SkynetClient, defaultSkynetPortalUrl } from "./index";
 
+const mock = new MockAdapter(axios);
+
 const portalUrl = defaultSkynetPortalUrl;
+const hnsLink = "doesn";
```
interstella-5555: let's use foo for more general test usage

```suggestion
const hnsLink = "foo";
```

src/download.test.js
```diff
@@ -55,3 +62,29 @@ describe("open", () => {
...
+
+      expect(mock.history.get.length).toBe(i);
+
+      expect(windowOpen).toHaveBeenCalledTimes(i);
+      expect(windowOpen).toHaveBeenCalledWith(`${portalUrl}/${skylink}`, "_blank");
+
+      i++;
+    }
+
+    mock.resetHistory();
```
interstella-5555: It's better to always reset mock before the assertions because in case of the failure, the reset might not be executed and subsequent tests will report fake success/failures. It might not be a case here but it's a good practice. Also it's a common practice in js to use function for looping array instead of an actual loop for increased safety of leaking variables out - again, not a case here but a good practice.

```suggestion
    validHnsLinkVariations.forEach((input, index) => {
      mock.resetHistory();
      
      await client.openHns(input);

      expect(mock.history.get.length).toBe(1);

      expect(windowOpen).toHaveBeenCalledTimes(1);
      expect(windowOpen).toHaveBeenCalledWith(`${portalUrl}/${skylink}`, "_blank");
    }
```

src/utils.js
interstella-5555: Then instead of arguments, we should use rest parameters. Usage of magic `arguments` is considered a bad practice due to it's magical nature :)

```suggestion
export function makeUrl(...args) {
  return args.reduce(function (acc, cur) {
    return urljoin(acc, cur);
  });
```

src/utils.js
interstella-5555: We don't have those documented anywhere and `uriHandshakeResolverPrefix` is not used anyway. Can we get some clarification on that @DavidVorick ?

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
-  it("should return single skylink on success", async () => {
-    const data = await client.uploadDirectory(directory, filename);
+  it("should send register onUploadProgress callback if defined", async () => {
+    const data = await client.uploadDirectory(directory, filename, { onUploadProgress: jest.fn() });
+
+    expect(mock.history.post.length).toBe(1);
+    const request = mock.history.post[0];
+    mock.resetHistory();
+
+    expect(request.onUploadProgress).toEqual(expect.any(Function));
```
interstella-5555, line 147: you should `expect` the function that you assigned to `onUploadProgress` on the first line, not just any function

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
 
-    expect(axios.post).toHaveBeenCalledWith(
-      `${portalUrl}/skynet/skyfile?filename=${filename}`,
-      expect.any(FormData), // TODO: Inspect data contents.
-      undefined
-    );
-  });
+    expect(mock.history.post.length).toBe(1);
+    const request = mock.history.post[0];
+    mock.resetHistory();
```
interstella-5555: reset history before the tests, not during (covered "why" in a different comment), same applies to other tests

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
+
+    expect(request.auth).toEqual({ username: "", password: "foobar" });
+    await compareFormData(request.data, [["file", "foo", filename]]);
+
+    expect(data).toEqual({ skylink });
+  });
+
+  it("should send custom user agent if defined", async () => {
+    const client2 = new SkynetClient(portalUrl, { customUserAgent: "Sia-Agent" });
```
interstella-5555: any reason to call it `client2` and not `client`?

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
+
+    // Should use client's user agent.
+
+    let data = await client2.upload(file);
```
interstella-5555:
```suggestion
    const data = await client2.upload(file);
```

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
+
+    expect(mock.history.post.length).toBe(1);
+    let request = mock.history.post[0];
```
interstella-5555:
```suggestion
    const request = mock.history.post[0];
```

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
+    mock.resetHistory();
+
+    expect(request.headers["User-Agent"]).toEqual("Sia-Agent");
+    // Check that other headers weren't altered.
+    expect(request.headers["Content-Type"]).toEqual("application/x-www-form-urlencoded");
+    await compareFormData(request.data, [["file", "foo", filename]]);
+
+    expect(data).toEqual({ skylink });
+
+    // Should use user agent set in options to function.
```
interstella-5555: that's a separate test

src/upload.test.js
```diff
@@ -1,97 +1,155 @@
...
+    // Use replyOnce to catch a single request with the new URL.
+    mock.onPost(url).replyOnce(200, { skylink: skylink });
+
+    const data = await client.upload(file, { onUploadProgress: jest.fn() });
+
+    expect(mock.history.post.length).toBe(1);
+    const request = mock.history.post[0];
+    mock.resetHistory();
+
+    expect(request.onUploadProgress).toEqual(expect.any(Function));
```
interstella-5555, line 50: expect the function from first line

interstella-5555 (changes requested):

package.json
mrcnski: Yeah, the tests were interfering with each other otherwise. For example I use `const request = mock.history.post[0];` to get the latest request made and it only works if one test runs at a time. There might be a way around this now that I think about it, though. I'll work on this.

src/upload.test.js
mrcnski: Doesn't work:

```
    expect(received).toEqual(expected) // deep equality

    Expected: [Function mockConstructor]
    Received: [Function anonymous]

      48 |     mock.resetHistory();
      49 | 
    > 50 |     expect(request.onUploadProgress).toEqual(jest.fn());
         |                                      ^
```

src/upload.test.js
mrcnski: See other comment

src/upload.test.js
interstella-5555: Because you need to store result of `jest.fn()` as a variable and compare with that variable. Otherwise, these are two different objects.

package.json
mrcnski: So I was thinking we could pass a query parameter like `?test=testName` to identify each request but it would make the tests more complicated. In band should be fine for now since the tests are quick to run.

src/upload.test.js
mrcnski: I tried that as well, same error

src/download.test.js
mrcnski: `forEach` doesn't work if you pass it an `async` function, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/download.js
mrcnski: That looks better to me too

src/upload.test.js
interstella-5555: It means the test detected a real issue then.

package.json
interstella-5555: That sounds like you leak the mocks somewhere. It shouldn't be an issue and can result in false-positives or false-negatives.

mrcnski: No, the problem is that the history gets scrambled with multiple requests running simultaneously, and then the test doesn't know which one to examine. Having one test run at a time and then erasing the history works fine. Open to suggestions though.

src/upload.test.js
mrcnski: Oh, it's because we modify `onUploadProgress` in `executeRequest`. Honestly it's enough to check here that `request.onUploadProgress` is defined and equal to any function.

mrcnski:
Just added some more changes based on my testing.

There is still an issue with redirects causing downloads to open in the same tab. I'll still need to test downloads once that is fixed. If that works I will need to also change the download tests.

Edit: have tested downloads with Ivo's [redirect fix MR](https://gitlab.com/NebulousLabs/Sia/-/merge_requests/4773) and they work.

src/index.test.js
```diff
@@ -5,12 +5,27 @@ describe("SkynetClient", () => {
...
     // Download
-    expect(client).toHaveProperty("download");
-    expect(client).toHaveProperty("getDownloadUrl");
-    expect(client).toHaveProperty("open");
+    expect(client).toHaveProperty("downloadFile");
+    expect(client).toHaveProperty("downloadFileHns");
+    expect(client).toHaveProperty("getHnsUrl");
+    expect(client).toHaveProperty("getHnsresUrl");
+    expect(client).toHaveProperty("getSkylinkUrl");
+    expect(client).toHaveProperty("metadata");
```
interstella-5555: this should be `getMetadata`

src/index.test.js
```diff
@@ -5,12 +5,27 @@ describe("SkynetClient", () => {
...
+    expect(client).toHaveProperty("getSkykeyById");
+    expect(client).toHaveProperty("getSkykeyByName");
+    expect(client).toHaveProperty("getSkykeys");
 
     // Upload
-    expect(client).toHaveProperty("upload");
+    expect(client).toHaveProperty("uploadFile");
+    expect(client).toHaveProperty("uploadFileRequest");
     expect(client).toHaveProperty("uploadDirectory");
+    expect(client).toHaveProperty("uploadDirectoryRequest");
```
interstella-5555, line 29: why do you need to separate the request methods?

webpack.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  mode: "development",
+  output: {
+    library: "skynet-js",
+    libraryTarget: "umd",
+    umdNamedDefine: true,
+  },
+  target: "web",
+};
```
interstella-5555: why is this needed?

src/index.test.js
mrcnski: It's a usability impediment to have to append `.skylink` to every upload call. Since that is the desired result 99% of the time we have this as the main API as well a version of the method that returns the full request (with merkleroot and bitfield).

interstella-5555 (approved):

PR #55, opened 2020-08-06
Update readme to point to the SDK docs

### LINKED ISSUE #39: Include the new functions in usage examples
FLW: We should include the new functions in README.md examples

_Originally posted by @kwypchlo in https://github.com/NebulousLabs/skynet-js/pull/37#issuecomment-660221457_

mrcnski:
Closes #39


interstella-5555 (changes requested):
Add examples to `open` `getDownloadUrl` and `parseSkylink`.

Also, it's weird to me that some functions are documented here and some in the api docs - I think all of them should be moved to API docs.

mrcnski:
> Also, it's weird to me that some functions are documented here and some in the api docs - I think all of them should be moved to API docs.

Yeah, I think so, too. I don't know yet how I'll do it without adding bloat for the other SDKs, which don't have these features.

mrcnski:
Okay, added them to the SDK docs. Raised https://github.com/NebulousLabs/skynet-docs/pull/26

interstella-5555 (approved):

PR #61, opened 2020-08-17
Download should stay in the same tab instead of open a new one

mrcnski:


Delivator:
Haven't tested but im 100% sure this will just open the skylink in the current window and if it's a file the browser can display, it won't start the download.

One way to start a download would be like this

```javascript
const downloadElement = document.createElement("a");
downloadElement.setAttribute("href", url);
downloadElement.setAttribute("download", true);
downloadElement.click();
```

But admittedly it's a bit hacky and I'm not sure how well it works on other browsers than chromium.

mrcnski:
Thanks a ton, this has had me stumped. I'm not a web guy at all lol

mrcnski:
I'm wondering if the download element shouldn't be deleted after it's clicked (assuming we go with this solution)

Delivator:
I'm not 100% sure on that but this way the element shouldn't even be added to the DOM.

Normally you would use `document.body.appendChild(element);` to add an element to the DOM.
Then you could use `document.body.removeChild(element);` or `element.remove();` but this won't do anything here.

Maybe `downloadElement = null;` but JS garbage collection could take care of it because the element is only created when the function is called.

mrcnski:
Thanks @Delivator, looks like you're right:

https://stackoverflow.com/a/1847289/6085242

src/download.js
```diff
@@ -11,7 +11,10 @@ SkynetClient.prototype.download = function (skylink, customOptions = {}) {
   const opts = { ...defaultDownloadOptions, ...customOptions, download: true };
   const url = this.getDownloadUrl(skylink, opts);
 
-  window.open(url, "_blank");
+  const downloadElement = document.createElement("a");
+  downloadElement.setAttribute("href", url);
+  downloadElement.setAttribute("download", true);
+  downloadElement.click();
```
interstella-5555, line 17: This should do the trick. Make sure that the url has `attachment=true` appended (setting `{ download: true }` in options passed to `getDownloadUrl ` is supposed to do that) - this will force siad to return the download content disposition that should force the browser to download.

```suggestion
  window.location = url;
```

interstella-5555 (changes requested):

src/download.js
interstella-5555: If that still doesn't work even after you confirm that `attachment=true` param is set in the url, check the content disposition on response headers on portal and on siad to verify that those are working correctly. The workaround with creating new element will not be needed to fix this.

Delivator: Yeah forgot about that option. That's way better

mrcnski:
Closing in favor of https://github.com/NebulousLabs/skynet-js/pull/48. I fixed downloads there and did manual testing. With that and this MR https://gitlab.com/NebulousLabs/Sia/-/merge_requests/4773 the issue will be fixed.

PR #76, opened 2020-09-09
Support downloading skylinks with paths and query params

### LINKED ISSUE #57: Set default maxBodyLength to Infinity
We should set `maxBodyLength` and `maxContentLength` to `Infinity` when making axios calls. See https://github.com/NebulousLabs/nodejs-skynet/issues/49

mrcnski:
See https://github.com/NebulousLabs/skynet-docs/issues/34.


src/download.js
```diff
@@ -28,34 +28,50 @@ const defaultResolveHnsOptions = {
  * @param {string} skylink - 46 character skylink.
  * @param {Object} [customOptions={}] - Additional settings that can optionally be set.
  * @param {string} [customOptions.endpointPath="/"] - The relative URL path of the portal endpoint to contact.
+ * @param {string} [customOptions.path] - An array of path elements to append to the skylink. Each element will be URI-encoded (e.g. "?" -> "%3F") so make sure it is not already encoded. Then the elements are joined to full the full path, e.g. `dir1/dir2/file`.
```
mrcnski:
```suggestion
 * @param {string} [customOptions.path] - An array of path elements to append to the skylink. Each element will be URI-encoded (e.g. "?" -> "%3F") so make sure it is not already encoded. Then the elements are joined to form the full path, e.g. `dir1/dir2/file`.
```

src/utils.js
```diff
@@ -82,7 +90,12 @@ export function parseSkylink(skylink) {
   // example: https://siasky.net/XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg
   const parsed = parse(skylink);
   const matchPathname = parsed.pathname.match(SKYLINK_PATHNAME_REGEX);
-  if (matchPathname) return matchPathname[SKYLINK_REGEXP_MATCH_POSITION];
+  if (matchPathname) {
+    const query = parsed.query;
+    // const hash = parsed.hash ? `#${parsed.hash}` : "";
```
mrcnski: Remove this comment

mrcnski:
@kwypchlo Pls review

src/utils.test.ts
```diff
@@ -52,18 +60,30 @@ describe("trimUriPrefix", () => {
...
+      [`${portalUrl}/${skylink}?foo=bar`, "?foo=bar"],
+      [`${portalUrl}/${skylink}/?foo=bar`, "/?foo=bar"],
+      [`${portalUrl}/${skylink}#foobar`, "#foobar"],
+      [`${portalUrl}/${skylink}/#foobar`, "/#foobar"],
     ];
 
-    validSkylinkVariations.forEach((input) => {
-      expect(parseSkylink(input)).toEqual(skylink);
+    validSkylinkVariations.forEach(([fullSkylink, path]) => {
+      expect(parseSkylink(fullSkylink)).toEqual(`${skylink}${path}`);
```
interstella-5555: `parseSkylink` should return just the raw skylink, without the path, otherwise it's a breaking change and we will need another function that does that same thing anyway since I have my use cases for that

I would suggest adding a function argument like `includePath` (that's false by default) that would return the skylink with the path.

Also I don't think we should be adding query parameters or hash values. Those are not part of the skylinks.

interstella-5555 (changes requested):

src/utils.test.ts
mrcnski: Okay I've addressed your points

src/utils.ts
```diff
@@ -63,17 +67,24 @@ export function makeUrl(...args: string[]): string {
...
+ */
+export function parseSkylink(skylink: string, includePath = false): string {
+  if (typeof skylink !== "string") {
+    throw new Error(`Skylink has to be a string, ${typeof skylink} provided`);
+  }
 
   // check for direct skylink match
   const matchDirect = skylink.match(SKYLINK_DIRECT_REGEX);
-  if (matchDirect) return matchDirect[SKYLINK_REGEXP_MATCH_POSITION];
+  if (matchDirect) return matchDirect[1];
```
interstella-5555: Why did you remove the const here ? It was added for the code to be a little more verbose.

src/download.ts
```diff
@@ -21,43 +21,71 @@ const defaultResolveHnsOptions = {
...
-  const query = opts.download ? { attachment: true } : {};
+  const query = opts.query ?? {};
+  if (opts.download) {
+    query.attachment = true;
+  }
+
+  // URL-encode the path.
+  let path = "";
+  if (opts.path) {
+    if (typeof opts.path != "string") {
```
interstella-5555:
```suggestion
    if (typeof opts.path !== "string") {
```

src/download.ts
```diff
@@ -21,43 +21,71 @@ const defaultResolveHnsOptions = {
...
+      throw new Error(`opts.path has to be a string, ${typeof opts.path} provided`);
+    }
+    // Encode each element of the path separately and join them.
+    path = opts.path
+      .split("/")
+      .map((element: string) => encodeURIComponent(element))
+      .join("/");
```
interstella-5555, line 85: shouldn't we call `encodeURI` on the whole url instead in makeUrl method ?

src/upload.test.ts
```diff
@@ -107,6 +107,21 @@ describe("uploadFile", () => {
...
   });
+
+  it("Should send custom query parameters if provided", async () => {
+    // Create a client with a unique portal so we don't accidentally hit another
+    // request mocker.
+    const portalUrl = "https://portal.net";
+    const url = `${portalUrl}/skynet/skyfile`;
+    const client = new SkynetClient(portalUrl, { customUserAgent: "Sia-Agent" });
+
+    mock.onPost(`${url}?file=test`).replyOnce(200, { skylink: skylink });
```
interstella-5555:
```suggestion
    mock.onPost(`${url}?file=test`).replyOnce(200, { skylink });
```

src/utils.test.ts
```diff
@@ -32,6 +37,9 @@ describe("makeUrl", () => {
     expect(makeUrl(portalUrl, "/", skylink)).toEqual(`${portalUrl}/${skylink}`);
     expect(makeUrl(portalUrl, "/skynet", skylink)).toEqual(`${portalUrl}/skynet/${skylink}`);
     expect(makeUrl(portalUrl, "//skynet/", skylink)).toEqual(`${portalUrl}/skynet/${skylink}`);
+    expect(makeUrl(portalUrl, "/skynet/", `${skylink}?foo=bar`)).toEqual(`${portalUrl}/skynet/${skylink}?foo=bar`);
+    expect(makeUrl(portalUrl, `${skylink}/?foo=bar`)).toEqual(`${portalUrl}/${skylink}?foo=bar`);
```
interstella-5555, line 52: shouldn't `?` be encoded in this example?

src/utils.test.ts
```diff
@@ -51,19 +59,31 @@ describe("trimUriPrefix", () => {
...
+      [`${skylink}/foo/bar`, false, ""],
+      [`${skylink}/foo/bar`, true, "/foo/bar"],
+      [`${skylink}#foobar`, true, ""],
+      [`sia:${skylink}`, true, ""],
+      [`sia://${skylink}`, true, ""],
+      [`${portalUrl}/${skylink}`, true, ""],
+      [`${portalUrl}/${skylink}/`, true, "/"],
+      [`${portalUrl}/${skylink}//`, true, "//"],
+      [`${portalUrl}/${skylink}?`, true, ""],
+      [`${portalUrl}/${skylink}/foo/bar`, false, ""],
```
interstella-5555: there are only two cases where you test this with `false` - there should be equal number of test cases for `false` and `true`

src/utils.ts
```diff
@@ -85,8 +96,10 @@ export function parseSkylink(skylink: string): string {
 
   // pass empty object as second param to disable using location as base url when parsing in browser
   const parsed = parse(skylink, {});
-  const matchPathname = parsed.pathname.match(SKYLINK_PATHNAME_REGEX);
-  if (matchPathname) return matchPathname[SKYLINK_REGEXP_MATCH_POSITION];
+  const matchIndirect = parsed.pathname.match(SKYLINK_PATHNAME_REGEX);
+  if (matchIndirect) {
+    return includePath ? matchIndirect[1] : matchIndirect[2];
```
interstella-5555: Please use constants for the key position `1` and `2` to make the code more descriptive

src/utils.test.ts
```diff
@@ -51,19 +59,31 @@ describe("trimUriPrefix", () => {
...
+      [`${portalUrl}/${skylink}/foo%3Fbar`, true, "/foo%3Fbar"],
+      [`${portalUrl}/${skylink}/foo/bar?foo=bar`, true, "/foo/bar"],
+      [`${portalUrl}/${skylink}?foo=bar`, true, ""],
+      [`${portalUrl}/${skylink}#foobar`, true, ""],
     ];
 
-    validSkylinkVariations.forEach((input) => {
-      expect(parseSkylink(input)).toEqual(skylink);
+    validSkylinkVariations.forEach(([fullSkylink, includePath, path]) => {
+      expect(parseSkylink(fullSkylink, includePath)).toEqual(`${skylink}${path}`);
```
interstella-5555: there are couple of missing cases, I think you should do a complete matrix where arguments are:

```
start with one of [ no-prefix | prefix | portalUrl ]
then add [ skylink ]
the add one of [ empty path | single slash | /foo/bar | /foo/bar/ ]
then add one of [ empty query | just ? | ?foo=bar ]
then add one of [ empty hash | just # | #baz ]
then choose includePath [ false | true ]
```

That would make `3 * 1 * 4 * 3 * 3 * 2 = 216` cases. Those don't need to be explicitly defined, should be generated automatically. You would expect just the skylink for all the cases with `includePath` to be equal to `skylink` and when `includePath` is set to `true` it would equal to `${skylink}${path}`.

I know that's a lot but I am not sure about the change in the regexp and we use that function in our portal code so I want to make sure that we will not throw an exception and break portal.

src/utils.ts
```diff
@@ -63,17 +67,24 @@ export function makeUrl(...args: string[]): string {
...
 }
 
-const SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
+// Allow ?, /, and # to end the hash portion of a skylink.
+const SKYLINK_BOUNDARY = "[/]";
+const SKYLINK_MATCHER = `([a-zA-Z0-9_-]{46})`;
 const SKYLINK_DIRECT_REGEX = new RegExp(`^${SKYLINK_MATCHER}$`);
-const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?${SKYLINK_MATCHER}([/?].*)?$`);
-const SKYLINK_REGEXP_MATCH_POSITION = 1;
+const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?(${SKYLINK_MATCHER}(${SKYLINK_BOUNDARY}+.*)?)$`);
```
interstella-5555: this looks like it should break on `/${skylink}?foo=bar` - I am not sure why it works since:

- this is the regexp `^/?(${SKYLINK_MATCHER}(${SKYLINK_BOUNDARY}+.*)?)$`
- this part `(${SKYLINK_BOUNDARY}+.*)?` is optional and it will not match in this case because it doesn't start with `/`
- the whole skylink should not be matched https://regex101.com/r/mN2W1k/1

src/download.test.ts
```diff
@@ -100,7 +155,7 @@ describe("getMetadata", () => {
   it("should fail quietly when skynet headers not present", () => {
     const headers = { "skynet-skylink": skylink };
 
-    validSkylinkVariations.forEach(async (input) => {
+    validSkylinkVariations.forEach(async ([input, _1, _2]) => {
```
interstella-5555: what is `_1` and `_2` ? even ts complains

Screenshot 2020-10-23 at 12 48 59

src/download.test.ts
```diff
@@ -29,50 +40,94 @@ Object.defineProperty(window, "location", {
...
+
+      let expectedUrl2 = `${expectedUrl}${path}${attachment}`;
+      if (path.startsWith("#")) {
+        // Hash should come after query.
+        expectedUrl2 = `${expectedUrl}${attachment}${path}`;
+      }
+      // Change ?attachment=true to &attachment=true if need be.
+      if ((expectedUrl2.match(/\?/g) || []).length > 1) {
+        expectedUrl2 = expectedUrl2.replace(attachment, "&attachment=true");
+      }
```
interstella-5555: I think that's too complicated logic for unit test to have, we should probably just stick with writing down the urls manually for each test case. Having regex and replace etc in a unit test sounds like that test could use a unit test to test that haha.

interstella-5555 (changes requested):

src/utils.ts
mrcnski: I didn't see the point of it, and I now use `1` and `2` for the skylink, and the skylink with path. I could have added a new const for `2` but I didn't see a reason for it.

src/utils.ts
mrcnski: I don't think it's necessary. Definitely makes the code harder to read rather than vice versa.

src/download.test.ts
mrcnski: How do I name unused parameters?

src/utils.ts
interstella-5555: Not at all, now you need to read the regex to know what 1 and 2 mean. Instead if you assign 1 and 2 to a variable, it will be self documenting.

src/download.test.ts
interstella-5555: Just don't add them.

src/utils.ts
mrcnski: I reworked this regex. No need to worry about `?foo=bar`, that is not part of `pathname`.

src/download.ts
mrcnski: Added a comment.

src/utils.test.ts
mrcnski: No, we'll encode paths passed in through `opts.path` since files on Skynet are allowed to contain ? etc. Otherwise the URL is treated as a URL. See the linked discussion.

src/download.test.ts
mrcnski: I think it is ok here. It tests our URL-forming code without using the URL-forming library so it is like an extra check, using different logic.

mrcnski:
Ready for another review @kwypchlo. I still need to add tests for getting the subdomain skylink from getSkylinkUrl, I'll do that if everything else looks good.

src/download.test.ts
```diff
@@ -42,23 +35,20 @@ Object.defineProperty(window, "location", {
...
-      expect(url).toEqual(expectedUrl2);
-      expect(mockLocationAssign).toHaveBeenCalledWith(expectedUrl2);
-    });
+  it.each(validSkylinkVariations)("should download with a skylink url with attachment set", (fullSkylink) => {
+    mockLocationAssign.mockClear();
+    const url = client.downloadFile(fullSkylink);
+
+    const path = extractNonSkylinkPath(fullSkylink, skylink);
+
+    let expectedUrl2 = `${expectedUrl}${path}${attachment}`;
```
interstella-5555: just call it `fullExpectedUrl`, `expected` or something like that, adding an number to a variable because the name is already taken is not really descriptive

src/download.ts
```diff
@@ -88,13 +88,24 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
   // TODO: fix subdomain + includePath
   let url;
   if (opts.subdomain) {
-    const skylinkPath = parseSkylink(skylink, { getPath: true });
-    skylink = parseSkylink(skylink);
+    // Get the path from the skylink.
+    const skylinkPath = parseSkylink(skylinkStr, { getPath: true });
```
interstella-5555: I would call `getPath` -> `onlyPath`, I actually had to look up what is it doing otherwise, wasn't really intuitive for me

interstella-5555 (changes requested):
all tests that have `it.each` should have `%s` in their name that interpolates the input string so the tests have unique names, otherwise we end up with 300 tests with the same name on each of such case

`it.each(invalidCases)("should return null on invalid case %s", (fullSkylink) => {`

interstella-5555 (approved):

PR #104, opened 2020-10-14
SkyDB functions

peterjan:


src/utils.ts
```diff
@@ -101,3 +101,7 @@ export function trimUriPrefix(str: string, prefix: string): string {
   }
   return str;
 }
+
+export function randomNumber(low: number, high: number): number {
```
m-cat, line 105: There's already a random number function

src/crypto.ts
```diff
@@ -0,0 +1,11 @@
+import * as blake2 from "blake2";
+
+// hashAll takes a list of string parameters and hashes them using the 'blake2b'
+// hasing function. The return value is a hex representation of that hash.
+export function hashAll(...args: string[]): string {
```
mrcnski: You should add a test to make sure we get the same expected values as in Sia.

src/crypto.ts
```diff
@@ -1,11 +1,69 @@
...
+      encodeNumber(fileID.version),
+      encodeString(fileID.applicationID),
+      encodeNumber(fileID.fileType),
+      encodeString(fileID.filename)
+    )
+  ).toString("hex");
+}
+
+// asciiToUint8Array converts a string to a uint8 array
+function asciiToUint8Array(str: string): Uint8Array {
```
mrcnski: Should move this back to `util.ts`. 

Also I would rename it as JS strings aren't ASCII (UTF-16 IIRC), `stringToAsciiUint8Array` maybe.

peterjan: I renamed but have not moved it. Because I changed the implementation, I only wanted my new code to use that function, instead of potentially breaking your encryption code. It is not exported so we can (re)move this at any time. I would rather keep it here until the SDK is out if everything works the way it is now.

src/crypto.ts
```diff
@@ -0,0 +1,69 @@
...
+      encodeString(fileID.applicationID),
+      encodeNumber(fileID.fileType),
+      encodeString(fileID.filename)
+    )
+  ).toString("hex");
+}
+
+// stringToUint8Array converts a string to a uint8 array
+function stringToUint8Array(str: string): Uint8Array {
+  return Uint8Array.from(Buffer.from(str));
```
mrcnski: `Buffer` doesn't work natively in the browser. You either need to use the string to uint8array function I have or use the `buffer` library, see my branch.

src/crypto.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// hexToUint8Array converts a hex encoded string to a uint8 array
+function hexToUint8Array(str: string) {
+  return new Uint8Array(str.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
+}
+
+// encodeNumber converts the given number into a uint8 array
+function encodeNumber(num: number): Uint8Array {
```
mrcnski:
```suggestion
function encodeUint8(num: number): Uint8Array {
```

src/crypto.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+  const encoded = new Uint8Array(8);
+  encoded[0] = num;
+  return encoded;
+}
+
+// encodeNumber converts the given string into a uint8 array
+function encodeString(str: string): Uint8Array {
+  const encoded = new Uint8Array(8 + str.length);
+  encoded.set(encodeNumber(str.length));
```
mrcnski: Looks like this function only works for strings up to length 255, have we confirmed that's always the case for all the strings we encode?

src/crypto.ts
```diff
@@ -0,0 +1,69 @@
...
+  for (let i = 0; i < args.length; i++) {
+    blake.blake2bUpdate(h, args[i]);
+  }
+  return blake.blake2bFinal(h);
+}
+
+// HashRegistryValue hashes the given registry value and returns it as a hex
+// encoded string
+export function HashRegistryValue(registryValue: RegistryValue): string {
+  return Buffer.from(
```
mrcnski: If you're using Buffer then you need the `buffer` library. It is already a dependency on my branch.

src/registry.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+export async function lookupRegistry(
+  user: User,
+  fileID: FileID,
+  customOptions = {}
+): Promise<SignedRegistryValue | null> {
+  const opts = {
+    ...defaultRegistryOptions,
+    ...customOptions,
```
mrcnski: the order of `customOptions` and `this.customOptions` should be reversed

mrcnski:
Please make sure every object that needs to be exported is present in `index.ts`

mrcnski:
Just add the buffer library and import it in every file that uses Buffer:
import { Buffer } from "buffer";

src/skydb.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  public constructor(protected publicKey: pki.ed25519.NativeBuffer, protected secretKey: pki.ed25519.NativeBuffer) {
+    this.id = publicKey.toString("hex");
+  }
+
+  // New takes a username and password and generates a key pair representing
+  // that user, it returns a new User object.
+  //
+  // NOTE: username should be the user's email address as ideally it's unique
+  public static New(username: string, password: string): User {
```
mrcnski: replace `New` with `constructor`. In JS we do `new User()` instead of `User.New`. Also elsewhere.

src/registry.ts
mrcnski: Also elsewhere

src/upload.test.ts
```diff
@@ -18,12 +18,12 @@ describe("uploadFile", () => {
   let mock: MockAdapter;
 
   beforeEach(() => {
-    mock = new MockAdapter(axios);
-    mock.onPost(url).reply(200, { skylink: skylink });
+    mock = new MockAdapter(axios, { onNoMatch: "passthrough" });
```
mrcnski: Remember to revert this

mrcnski (changes requested):
Looks good for the most part. Let's just get the critical stuff fixed and the rest can go in a follow-up.

src/registry.ts
peterjan: thanks

src/crypto.ts
peterjan: yes I think so - but we should definitely double check

mrcnski (approved):

PR #105, opened 2020-10-14
Implemented getMetadata

dghelm:
Implemented getMetadata by calling getSkylinkUrl, doing a HEAD request on this, and then returning the JSON parse of the "skynet-file-metadata" header.


src/download.ts
```diff
@@ -70,8 +70,17 @@ export function getHnsresUrl(domain: string, customOptions = {}) {
 
 export async function getMetadata(skylink: string, customOptions = {}) {
   const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
+  const url = this.getSkylinkUrl(skylink, opts);
+
+  console.log(url);
```
interstella-5555: please remove `console.log` once this gets ready for merge

interstella-5555 (changes requested):
That's a great contribution! Please add a test case for this function though, we need to make sure it works.

mrcnski (changes requested):
Thanks, looks great! Just needs a quick test.

dghelm:
Test added. Console Log removed.

Function modified to return an empty object if no "skynet-file-metadata" present. Should this instead throw an error?

src/download.test.ts
```diff
@@ -118,6 +153,7 @@ describe("openFileHns", () => {
 
   beforeEach(() => {
     mock = new MockAdapter(axios);
+    const data = {};
```
mrcnski: Superfluous line?

src/download.ts
```diff
@@ -72,15 +72,22 @@ export async function getMetadata(skylink: string, customOptions = {}) {
...
-  console.log(url);
-
-  const response = await this.executeRequest({
-    ...opts,
-    method: "head",
-    url,
-  });
-
-  return JSON.parse(response.headers["skynet-file-metadata"]);
+  let skynetFileMetadata = {};
```
mrcnski: Minor point: you can save a couple lines by returning `skynetFileMetadata` from within the `try` block. No need to declare it here.

mrcnski (approved):
Awesome, looks great! I ran a manual test and confirmed that your code works 😁

src/download.test.ts
dghelm: Yup! Will remove.

src/download.ts
dghelm: Good catch.

mrcnski (approved):

interstella-5555 (approved):
Looks great!

PR #108, opened 2020-10-16
Timeout using Promise.Race

peterjan:


src/utils.ts
```diff
@@ -107,12 +107,15 @@ export function randomNumber(low: number, high: number): number {
...
 }
 
-export async function timeout(ms: number) {
-  return new Promise((resolve, reject) => {
-    setTimeout(() => {
-      resolve(null);
+export async function promiseTimeout(ms: number, promise: any) {
+  const timeout = new Promise((resolve, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
```
ro-tex: What does `clearTimeout` do in this case? If the timeout hasn't fired yet it would remove it but if it has does it make any difference?

src/utils.ts
```diff
@@ -107,12 +107,15 @@ export function randomNumber(low: number, high: number): number {
   return Math.random() * (high - low) + low;
 }
 
-export async function timeout(ms: number) {
-  return new Promise((resolve, reject) => {
-    setTimeout(() => {
-      resolve(null);
+export async function promiseTimeout(ms: number, promise: any) {
```
ro-tex: I'm not really sure about this but it might be nice to have the promise come first in order to mimic the `setTimeout` signature that all JS devs are so used to.

ro-tex (approved):
LGTM

mrcnski (approved):

mrcnski (approved):

PR #123, opened 2020-10-19
SkyDB Fixes

peterjan:
This PR allows encoding numbers higher than 255. It also uses SHA256 as digest in the key derivation function.

Note that it does not yet use a big int for the revision number, as we should. I believe Marcin uses https://github.com/GoogleChromeLabs/jsbi in the encryption branch, so if we use one it should be that one. Have not done that yet because it would need to be part of the interface I think, seeing as there are uncertainties still on what the interface is supposed to be I want to hold off on that.


src/crypto.ts
```diff
@@ -34,11 +34,12 @@ export function HashFileID(fileID: FileID): Uint8Array {
...
 function encodeUint8(num: number): Uint8Array {
-  if (num > 255) {
-    throw new Error("overflow");
-  }
   const encoded = new Uint8Array(8);
-  encoded[0] = num;
+  for (let index = 0; index < encoded.length; index++) {
+    const byte = num & 0xff;
+    encoded[index] = byte;
+    num = (num - byte) / 256;
```
mrcnski: I would just do `>> 8` here.

Also, does this match the encode function in Sia? We need to make sure all encoding is done the same

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
...
+    expect(existing).toBeNull();
+
+    // build the registry value
+    let revision = 0;
+    const skylink = "CABAB_1Dt0FJsxqsu_J4TodNCbCGvtFf1Uys_3EgzOlTcg";
+
+    // update the registry 256 times
+    for (revision; revision <= 260; revision += 10) {
+      if (revision && revision % 50 === 0) {
+        console.log(`Iteration ${revision}...`);
```
interstella-5555: remove console.log before merging that to master

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
...
+    console.log(`Uploading for userid ${user.id}`);
+
+    // verify there is no entry
+    let existing = null;
+    try {
+      existing = await promiseTimeout(client.lookupRegistry(user, fileID), 1e4);
+    } catch (error) {
+      // expected
+    }
+    expect(existing).toBeNull();
```
interstella-5555: use expect toThrow assertion https://jestjs.io/docs/en/expect

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
...
+
+    // build the registry value
+    let revision = 0;
+    const skylink = "CABAB_1Dt0FJsxqsu_J4TodNCbCGvtFf1Uys_3EgzOlTcg";
+
+    // update the registry 256 times
+    for (revision; revision <= 260; revision += 10) {
```
interstella-5555:
```suggestion
    const skylink = "CABAB_1Dt0FJsxqsu_J4TodNCbCGvtFf1Uys_3EgzOlTcg";

    // update the registry 256 times
    for (let revision = 0; revision <= 260; revision += 10) {
```

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
...
+
+      // update the registry entry
+      const updated = await client.updateRegistry(user, fileID, { value, signature });
+      expect(updated).toBe(true);
+    }
+
+    // verify the revision in the registry
+    existing = await client.lookupRegistry(user, fileID);
+    expect(existing.value.revision).toBeGreaterThan(255);
+    console.log(existing.value);
```
interstella-5555: drop console.log before merging

src/skydb.test.ts
```diff
@@ -20,6 +20,15 @@ describe("User", () => {
...
   });
+
+  it.skip("should take a considerable amount of time", async () => {
+    const start = new Date().getTime();
+    new User("john.doe@example.com", "supersecret");
+    const elapsed = new Date().getTime() - start;
+    expect(elapsed).toBeGreaterThan(10);
+    expect(elapsed).toBeLessThan(100);
+    console.log(`${elapsed}ms`);
+  });
```
interstella-5555: what does this test do?

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
...
+      const signature = user.sign({ message: HashRegistryValue(value) });
+
+      // update the registry entry
+      const updated = await client.updateRegistry(user, fileID, { value, signature });
+      expect(updated).toBe(true);
+    }
+
+    // verify the revision in the registry
+    existing = await client.lookupRegistry(user, fileID);
+    expect(existing.value.revision).toBeGreaterThan(255);
```
interstella-5555:
```suggestion
    const entry = await client.lookupRegistry(user, fileID);
    expect(entry.value.revision).toBeGreaterThan(255);
```

don't overwrite a variable if it's not necessary

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
     const buf = Buffer.from(parts[1], "base64");
     expect(buf.toString("ascii")).toEqual("thisistext");
   });
+
+  it("support rev number update higher than 255", async () => {
+    // create a random user
+    const rand = randomNumber(0, 1e6);
+    const user = new User(`john.doe+${rand}@gmail.com`, "test1234");
+    console.log(`Uploading for userid ${user.id}`);
```
interstella-5555: drop console.log before merging

src/siaskydev.test.ts
```diff
@@ -29,4 +31,48 @@ describe.skip("siasky.dev end to end", () => {
     const buf = Buffer.from(parts[1], "base64");
     expect(buf.toString("ascii")).toEqual("thisistext");
   });
+
+  it("support rev number update higher than 255", async () => {
+    // create a random user
+    const rand = randomNumber(0, 1e6);
+    const user = new User(`john.doe+${rand}@gmail.com`, "test1234");
```
interstella-5555:
```suggestion
    const user = new User("john.doe@gmail.com", "test1234");
```
you don't need random number there, it's isolated test case

interstella-5555 (changes requested):

src/siaskydev.test.ts
peterjan: Why? The `siaskydev.test.ts` file is skipped and marked as follows:

```
describe.skip("siasky.dev end to end", () => {
```

peterjan: I'll resolve all the `console.log` remarks and keep this open.

src/siaskydev.test.ts
peterjan: It doesn't really matter much if I randomize the user or the file id right? It's not really isolated either because the SkynetClient is pointed at `siasky.dev`. If anything I'd rather remove this test file altogether if it's an issue keeping it in.  I only wrote these functions to be able to quickly verify the working on siasky.dev instead of relying on the mock adapter.

peterjan: I need to randomize either the user or the file id.

src/skydb.test.ts
peterjan: quickly wrote that to verify how long it took to create a new user on my machine, 
was only looking for order of magnitudes here while playing with the key derivation function input
(like `numIterations` and digest function)

src/siaskydev.test.ts
peterjan: I tried that but it doesn't work for async functions of which you want to assert the result value. I also don't really care about the error thrown, wouldn't you normally use `toThrow` when you want to assert what the error?

peterjan: I would use, that's not it but the following's not working
`expect(async () => {await promiseTimeout(client.lookupRegistry(user, fileID), 1e4);}).toThrow('..my error')`
and I much rather use the return value of that function in the assert.

package.json
```diff
@@ -7,7 +7,7 @@
     "dist/*"
   ],
   "scripts": {
-    "test": "jest",
+    "test": "jest --detectOpenHandles",
```
interstella-5555: please revert that before merging to master

src/skydb.test.ts
interstella-5555: either remove that or make it a proper test before merging to master please

src/siaskydev.test.ts
interstella-5555: Right, I did not realise this is an e2e test. We should probably move it to a separate suite so it's not confused with the unit tests.

src/siaskydev.test.ts
interstella-5555: you can try a suggestion from https://github.com/facebook/jest/issues/5538
```javascript
await expect(asyncFunc()).rejects.toThrow();
```

interstella-5555: https://stackoverflow.com/a/47887098

peterjan: I tried, it does not work the way I want it to. I don't see why we would rewrite it in an awkward way only to avoid the `try-catch` in a test file. I looked it up and everything, it's not that I did not try, I just don't see the issue using `try-catch` here.

I much rather remove this entire file if it is cause for concern or something we do not want to merge per se. I only kept it in because it is useful to have for debugging purposes.

package.json
peterjan: Why? I reverted it but I originally kept it on purpose, not by mistake.

It helps to provide some context as to why you want something to be different. Especially since I am not familiar with having to writing javascript nor making pull requests in this repo.

interstella-5555: https://jestjs.io/docs/en/cli

> --detectOpenHandles
Attempt to collect and print open handles preventing Jest from exiting cleanly. Use this in cases where you need to use --forceExit in order for Jest to exit to potentially track down the reason. This implies --runInBand, making tests run serially. Implemented using async_hooks. This option has a significant performance penalty and should only be used for debugging.

Specifically
> Use this in cases where you need to use --forceExit in order for Jest to exit to potentially track down the reason.

Is this your use case? You should run this locally if so, fix the issue and run normally afterwards.

src/skydb.test.ts
peterjan: Not sure what you mean with "proper test". I assume document more? Not sure though, it's hardware dependant and I don't think it's worth the effort. That is also why I originally ended up skipping it and now decided to remove it altogether.

interstella-5555 (approved):

interstella-5555 (changes requested):
it seems that right now there is no unit test for this change though

peterjan:
@kwypchlo I updated the unit test to have a revision number which would reveal a mistake against the encoding. I am not sure how we can test the change to HMAC256, there is however a test that validates user creation. Do you think we need anything besides these unit tests?

peterjan:
an integration test might make more sense, I'm just not sure where to add such a test or what would constitute a good integration test at this point, we are still in the process of defining everything and everything is potentially up for changing

interstella-5555 (approved):

mrcnski (approved):

PR #126, opened 2020-10-20
Update doc links

mrcnski:


interstella-5555 (approved):

package.json
```diff
@@ -51,7 +51,7 @@
   "bugs": {
     "url": "https://github.com/NebulousLabs/skynet-js/issues"
   },
-  "homepage": "https://nebulouslabs.github.io/skynet-js/",
+  "homepage": "https://github.com/NebulousLabs/skynet-js",
```
interstella-5555, line 54: I think it is supposed to be either https://nebulouslabs.github.io/skynet-js or https://siasky.net/docs

mrcnski: I think it is correct as is:

> `homepage`: The url to the project homepage.

- https://docs.npmjs.com/files/package.json

PR #127, opened 2020-10-20
Add support for base32 skylinks and subdomains

### LINKED ISSUE #93: Base32 encoding and hns subdomains support
We introduced 2 new features recently:

### Base32 skylink subdomain access
You can reencode skylink with Base32 and use it as a subdomain to access your content on siasky.net:

- Regular skylink: https://siasky.net/XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg
- Base32 skylink in subdomain: https://bg06v2tidkir84hg0s1s4t97jaeoaa1jse1svrad657u070c9calq4g.siasky.net

Converter: https://convert-skylink.hns.siasky.net/ (conversion explained and documented on [GitHub](https://github.com/kwypchlo/base32))

We should add support to either return regular skylink url or the subdomain skylink url when returning skylink url in SDK. preferably regular skylink url by default and subdomain skylink url if some optional param is set to `true`.

We should also add a utility function that converts Base64 (regular) skylink (46 chars) to Base32 skylink. There are already decoder and encoder functions here https://github.com/kwypchlo/base32/blob/master/src/crypto.ts and usage example is https://github.com/kwypchlo/base32/blob/master/src/App.tsx#L24-L25 but I guess it could be one function in the SDK (instead of two).

### Handshake subdomain access
You can access your content bound to the hns domain using *.hns.siasky.net subdomain:

Regular hns: https://siasky.net/hns/convert-skylink/
New hns subdomain: https://convert-skylink.hns.siasky.net/

We should add a param to function that return hns url to either return regular one (default) or the one in subdomain (optional param set to `true`).

### LINKED ISSUE #143: parseSkylink should return null on not found
From https://github.com/NebulousLabs/skynet-js/pull/141#discussion_r512120269.

mrcnski:
Closes #93 
Closes #143

Note to self: Update docs to include the new options.


src/download.test.ts
```diff
@@ -6,7 +6,9 @@ import { SkynetClient, defaultSkynetPortalUrl } from "./index";
 const portalUrl = defaultSkynetPortalUrl;
 const hnsLink = "foo";
 const hnsUrl = `${portalUrl}/hns/${hnsLink}`;
+const hnsUrlSubdomain = `https://${hnsLink}.hns.siasky.net`;
 const hnsresUrl = `${portalUrl}/hnsres/${hnsLink}`;
+const hnsresUrlSubdomain = `https://${hnsLink}.hnsres.siasky.net`;
```
interstella-5555: there is no such thing - hnsres doesn't get a subdomain, only skylinks and hns

src/download.ts
```diff
@@ -23,7 +25,9 @@ const defaultResolveHnsOptions = {
  * Initiates a download of the content of the skylink within the browser.
  * @param {string} skylink - 46 character skylink.
  * @param {Object} [customOptions={}] - Additional settings that can optionally be set.
+ * @param {boolean} [customOptions.base32=false] - Whether to return a base32 skylink instead of base64.
```
interstella-5555: I don't think this param makes sense, does it?

src/download.ts
```diff
@@ -51,22 +56,35 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
   const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
   const query = opts.download ? { attachment: true } : {};
 
-  const url = makeUrl(this.portalUrl, opts.endpointPath, parseSkylink(skylink));
+  skylink = parseSkylink(skylink);
+  if (opts.base32) {
+    skylink = convertSkylinkToBase32(skylink);
+  }
```
interstella-5555:
```suggestion
  if (opts.subdomain) {
    skylink = convertSkylinkToBase32(skylink);
  }
```

src/download.ts
```diff
@@ -51,22 +56,35 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
...
+  const url = opts.subdomain
+    ? addSubdomain(this.portalUrl, skylink)
+    : makeUrl(this.portalUrl, opts.endpointPath, skylink);
```
interstella-5555, line 65: Can `makeUrl` handle the subdomain internally?

src/download.ts
```diff
@@ -51,22 +56,35 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
...
 
 export function getHnsUrl(this: SkynetClient, domain: string, customOptions: any = {}): string {
   const opts = { ...defaultDownloadHnsOptions, ...this.customOptions, ...customOptions };
   const query = opts.download ? { attachment: true } : {};
 
-  const url = makeUrl(this.portalUrl, opts.endpointPath, trimUriPrefix(domain, uriHandshakePrefix));
+  domain = trimUriPrefix(domain, uriHandshakePrefix);
+  const url = opts.subdomain
+    ? addSubdomain(addSubdomain(this.portalUrl, opts.endpointPath), domain)
+    : makeUrl(this.portalUrl, opts.endpointPath, domain);
```
interstella-5555: we should abstract out this url composing, also with the query params

src/download.ts
```diff
@@ -51,22 +56,35 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
...
 
 export function getHnsresUrl(this: SkynetClient, domain: string, customOptions: any = {}): string {
   const opts = { ...defaultResolveHnsOptions, ...this.customOptions, ...customOptions };
 
-  return makeUrl(this.portalUrl, opts.endpointPath, trimUriPrefix(domain, uriHandshakeResolverPrefix));
+  domain = trimUriPrefix(domain, uriHandshakeResolverPrefix);
+  const url = opts.subdomain
+    ? addSubdomain(addSubdomain(this.portalUrl, opts.endpointPath), domain)
+    : makeUrl(this.portalUrl, opts.endpointPath, domain);
+  return url;
```
interstella-5555: this should be reverted, hnsres is not included in the subdomain access

src/download.ts
```diff
@@ -106,6 +124,7 @@ export async function openFileHns(this: SkynetClient, domain: string, customOpti
  * @param {string} domain - Handshake resolver domain.
  * @param {Object} [customOptions={}] - Additional settings that can optionally be set.
  * @param {string} [customOptions.endpointPath="/hnsres"] - The relative URL path of the portal endpoint to contact.
+ * @param {boolean} [customOptions.subdomain=false] - Whether to return the final URL with the HNS domain as a subdomain.
```
interstella-5555: this is redundant here

src/utils.ts
```diff
@@ -10,12 +12,31 @@ export const uriHandshakePrefix = "hns:";
...
+  while (subdomain.startsWith("/")) {
+    subdomain = subdomain.slice(1);
+  }
+  if (url.includes("https://")) {
+    return url.replace("https://", `https://${subdomain}.`);
+  }
+  if (url.includes("http://")) {
+    return url.replace("http://", `http://${subdomain}.`);
+  }
+  return `${subdomain}.${url}`;
```
```suggestion
  const urlObj = new URL(url);
  urlObj.hostname = `${subdomain}.${urlObj.hostname}`;
  return urlObj.toString();
```

src/utils.ts
```diff
@@ -10,12 +12,31 @@ export const uriHandshakePrefix = "hns:";
 export const uriHandshakeResolverPrefix = "hnsres:";
 export const uriSkynetPrefix = "sia:";
 
+// TODO: Use a third-party library to make this more robust.
+export function addSubdomain(url: string, subdomain: string): string {
+  while (subdomain.startsWith("/")) {
+    subdomain = subdomain.slice(1);
+  }
```
interstella-5555: why would subdomain ever start with leading slash ?

interstella-5555 (changes requested):

src/download.ts
mrcnski: I don't think so, it takes a variable number of arguments so it can't take optional arguments. Don't think it's necessary either.

src/download.ts
mrcnski: What would that look like? Anyway, premature abstraction is not always the greatest thing.

src/download.test.ts
```diff
@@ -74,6 +76,16 @@ describe("getSkylinkUrl", () => {
...
     expect(url).toEqual(`${portalUrl}/skynet/skylink/${skylink}?attachment=true`);
   });
+
+  it("should convert base64 skylinks to base32", () => {
+    const expectedBase32 = "https://bg06v2tidkir84hg0s1s4t97jaeoaa1jse1svrad657u070c9calq4g.siasky.net";
+
+    validSkylinkVariations.forEach((input) => {
+      const url = client.getSkylinkUrl(input, { base32: true, subdomain: true });
+
+      expect(url).toEqual(`${expectedBase32}/`);
```
interstella-5555: This should equal to that

```suggestion
      expect(url).toEqual(expectedBase32);
```

src/download.test.ts
```diff
@@ -74,6 +76,16 @@ describe("getSkylinkUrl", () => {
 
     expect(url).toEqual(`${portalUrl}/skynet/skylink/${skylink}?attachment=true`);
   });
+
+  it("should convert base64 skylinks to base32", () => {
+    const expectedBase32 = "https://bg06v2tidkir84hg0s1s4t97jaeoaa1jse1svrad657u070c9calq4g.siasky.net";
+
+    validSkylinkVariations.forEach((input) => {
+      const url = client.getSkylinkUrl(input, { base32: true, subdomain: true });
```
interstella-5555: `base32` property should not be necessary here

```suggestion
      const url = client.getSkylinkUrl(input, { subdomain: true });
```

src/download.ts
```diff
@@ -51,22 +56,33 @@ export function getSkylinkUrl(this: SkynetClient, skylink: string, customOptions
...
   return addUrlQuery(url, query);
 }
 
 export function getHnsresUrl(this: SkynetClient, domain: string, customOptions: any = {}): string {
   const opts = { ...defaultResolveHnsOptions, ...this.customOptions, ...customOptions };
 
-  return makeUrl(this.portalUrl, opts.endpointPath, trimUriPrefix(domain, uriHandshakeResolverPrefix));
+  domain = trimUriPrefix(domain, uriHandshakeResolverPrefix);
+  const url = makeUrl(this.portalUrl, opts.endpointPath, domain);
+  return url;
```
interstella-5555: Let's revert to what was here before - I don't think any changes here are necessary

```suggestion
  return makeUrl(this.portalUrl, opts.endpointPath, trimUriPrefix(domain, uriHandshakeResolverPrefix));
```

src/utils.test.ts
```diff
@@ -67,6 +79,10 @@ describe("parseSkylink", () => {
     });
   });
 
+  it("should parse out base32 skylink from subdomain", () => {
+    expect(parseSkylink(portalUrlSubdomain, { subdomain: true })).toEqual(skylinkBase32);
```
interstella-5555: I would expect `parseSkylink` to work like this:

```suggestion
    expect(parseSkylink(portalUrlSubdomain)).toEqual(skylink);
```

src/utils.test.ts
```diff
@@ -7,10 +7,14 @@ import {
...
   uriHandshakeResolverPrefix,
   getFileMimeType,
+  convertSkylinkToBase32,
+  addSubdomain,
 } from "./utils";
 
 const portalUrl = defaultSkynetPortalUrl;
 const skylink = "XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg";
+const skylinkBase32 = "bg06v2tidkir84hg0s1s4t97jaeoaa1jse1svrad657u070c9calq4g";
+const portalUrlSubdomain = addSubdomain(portalUrl, skylinkBase32);
```
interstella-5555: this should be hardcoded string, in case `addSubdomain` malfunctions, we will not be able to pinpoint the issue here

```suggestion
const portalUrlSubdomain = `https://${skylinkBase32}.siasky.net`;
```

src/utils.ts
```diff
@@ -63,12 +77,15 @@ export function makeUrl(...args: string[]): string {
...
-const SKYLINK_DIRECT_REGEX = new RegExp(`^${SKYLINK_MATCHER}$`);
-const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?${SKYLINK_MATCHER}([/?].*)?$`);
-const SKYLINK_REGEXP_MATCH_POSITION = 1;
+export function parseSkylink(skylink: string, opts: any = {}): string {
+  let SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
+  if (opts.subdomain) {
+    SKYLINK_MATCHER = "([a-z0-9_-]{55}).";
+  }
+  const SKYLINK_DIRECT_REGEX = new RegExp(`^${SKYLINK_MATCHER}$`);
+  const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?${SKYLINK_MATCHER}([/?].*)?$`);
```
interstella-5555: move those consts outside of `parseSkylink`, otherwise those regexes will be recompiled every time this function is called

src/utils.ts
```diff
@@ -63,12 +77,15 @@ export function makeUrl(...args: string[]): string {
...
 
-const SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
-const SKYLINK_DIRECT_REGEX = new RegExp(`^${SKYLINK_MATCHER}$`);
-const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?${SKYLINK_MATCHER}([/?].*)?$`);
-const SKYLINK_REGEXP_MATCH_POSITION = 1;
+export function parseSkylink(skylink: string, opts: any = {}): string {
+  let SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
+  if (opts.subdomain) {
+    SKYLINK_MATCHER = "([a-z0-9_-]{55}).";
+  }
```
interstella-5555: this function should be intelligent enough to determine that by itself, you shouldn't need to pass "opts.subdomain"

interstella-5555 (changes requested):

src/utils.ts
mrcnski: We discussed that the function shouldn't try to be "smart", there could be e.g. non-skylink subdomains that happen to be 55 chars long.

src/utils.ts
```diff
@@ -77,35 +81,64 @@ export function makeUrl(...args: string[]): string {
   return args.reduce((acc, cur) => urljoin(acc, cur));
 }
 
-export function parseSkylink(skylink: string, opts: any = {}): string {
-  let SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
+const SKYLINK_MATCHER = "([a-zA-Z0-9_-]{46})";
+const SKYLINK_MATCHER_SUBDOMAIN = "([a-z0-9_-]{55})";
+const SKYLINK_DIRECT_REGEX = new RegExp(`^${SKYLINK_MATCHER}$`);
+// const SKYLINK_SUBDOMAIN_PATHNAME_REGEX = new RegExp(`^${SKYLINK_MATCHER}(`);
```
interstella-5555: remove

src/utils.ts
```diff
@@ -77,35 +81,64 @@ export function makeUrl(...args: string[]): string {
...
+const SKYLINK_PATHNAME_REGEX = new RegExp(`^/?${SKYLINK_MATCHER}(/.*)?$`);
+const SKYLINK_SUBDOMAIN_REGEX = new RegExp(`^/?${SKYLINK_MATCHER_SUBDOMAIN}(\\..*)?$`);
```
interstella-5555: I dont think the leading `/?` is valid in this case

interstella-5555 (changes requested):

interstella-5555:
Added some failing test cases to `parseSkylink` in https://github.com/NebulousLabs/skynet-js/pull/127/commits/e4617acde5a9484fc2485a2b16859dca62573b00

interstella-5555 (approved):

PR #128, opened 2020-10-21
Improve ts support

interstella-5555:
- document `this` with typescript on client methods
- clean up `CustomClientOptions`
- replace `windows.location = url` with `window.location.assign(url)` to make typescript and tests happy

TODO: @m-cat will update docs on client options

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -27,6 +27,11 @@ export class SkynetClient {
   portalUrl: string;
   customOptions: CustomClientOptions;
 
+  /**
+   * The Skynet Client which can be used to access Skynet.
+   * @param portalUrl The portal URL to use to access Skynet, if specified. To use the default portal while passing custom options, use ""
+   * @param customOptions Configuration for the client
+   */
   constructor(portalUrl: string = defaultPortalUrl(), customOptions: CustomClientOptions = {}) {
     this.portalUrl = portalUrl;
     this.customOptions = customOptions;
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosResponse, AxiosRequestConfig } from "axios";
+import axios, { AxiosResponse } from "axios";
 import { uploadFile, uploadDirectory, uploadDirectoryRequest, uploadFileRequest } from "./upload";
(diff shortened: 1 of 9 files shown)


src/client.ts
```diff
@@ -17,25 +17,17 @@ import { addUrlQuery, defaultPortalUrl, makeUrl } from "./utils";
 import { getFile, setFile } from "./skydb";
 import { lookupRegistry, updateRegistry } from "./registry";
 
+export type CustomClientOptions = {
+  APIKey?: string; // authentication password to use
```
mrcnski: Can we use JSDoc to document these parameters?

src/client.ts
```diff
@@ -17,25 +17,17 @@ import { addUrlQuery, defaultPortalUrl, makeUrl } from "./utils";
...
+
 export class SkynetClient {
   portalUrl: string;
-  customOptions: Record<string, unknown>;
+  customOptions: CustomClientOptions;
 
-  /**
-   * The Skynet Client which can be used to access Skynet.
-   * @constructor
-   * @param {string} [portalUrl="https://siasky.net"] - The portal URL to use to access Skynet, if specified. To use the default portal while passing custom options, use "".
```
mrcnski, line 27: Why remove all this documentation instead of just the irrelevant parameters?

mrcnski (approved):

mrcnski (changes requested):
Changes requested!

src/client.ts
interstella-5555: yeah sorry, my bad

src/client.ts
interstella-5555: jsdoc doesn't apply to typescript

interstella-5555: replaced `// comment` with `/** comment */` so they are preserved in the resulting d.ts file

src/client.ts
mrcnski: Should we indicate that these parameters are optional? (e.g. with brackets in the JSDoc or a `?` after the parameter name)

interstella-5555: yeah, wrapped back in brackets

mrcnski (approved):

Issue #129, opened 2020-10-21
requestFile Function


dghelm:
Currently, the SDK uses very similar ways of directing the browser to "download" or "open" files, both by calling the window object. 

As I work more with skynet-js, I'm wondering if "requesting" a file and returning the data of the Axios response (and perhaps skylink metadata) makes sense, especially for txt, json or other filetypes.

I'd propose `getFile` for this, but that overlaps with SkyDB function names, so perhaps `client.requestFile(skylink)`? 

I'd be happy to do the initial implementation.

mrcnski:
Yeah, this is definitely something we want and that is needed for the ecosystem. `requestFile` sounds good. `downloadFile` should stay the same since we are guaranteeing backwards compatibility for the SDK since v2 (for everything except the SkyDB functions which are an explicitly unstable API).

> I'd be happy to do the initial implementation.

That would be great! 🙂 Let us know if you have any questions.

PR #130, opened 2020-10-22
requestFile Implementation

dghelm:
Initial implementation for requestFile. No sanity checks for large files or anything else.

Addresses issue #129


mrcnski (approved):

kwypchlo:
I would wonder whether it should be called `readFile` or `getFileContent` instead. Probably `getFileContent` is a better idea and would sound more cohesive with `getMetadata`.

interstella-5555 (commented):
I would wonder whether it should be called `readFile` or `getFileContent` instead. Probably `getFileContent` is a better idea and would sound more cohesive with `getMetadata`. 

edit: or maybe `getContent` but I would argue it's too vague and generic

dghelm:
I agree that `Request` is a little clunky here. My concern with `getFileContent` is that `getFile` is using SkyDB (at least with its initial SDK), so I think that gets confusing, but I think that's just an issue with the SkyDB method name more than anything in the Download methods.

I think with the current naming structure, `readFile` or `readSkylink` make sense, as well as `getSkylinkContents` since there is a `getSkylinkUrl`. That said (and this is obviously not a conversation for a pull request), there may need to be a standardized naming structure (even if just aliases for existing methods) that allow the developer to always know what data space they're dealing with (skylinks, metadata, files, db entries, etc).

mrcnski:
> That said (and this is obviously not a conversation for a pull request), there may need to be a standardized naming structure (even if just aliases for existing methods) that allow the developer to always know what data space they're dealing with (skylinks, metadata, files, db entries, etc).

We started doing this with SkyDB and the Registry (e.g. `client.db.getJSON` and `client.registry.getEntry`)

With this in mind I think `getFileContent` would be a good name, especially since we ended up removing `getFile` in the final version of the SkyDB API.

MSevey:
Comments, please add comments and descriptions to the new functions :-)

dghelm:
Okay I think it should be good to go.

mrcnski (dismissed):

interstella-5555 (dismissed):
I am going to approve it but we should take a look at our SDK methods. If we would want to be consistent with the naming, we already have `getSkylinkUrl` so this new method should probably be `getSkylinkContent`. Otherwise `getSkylinkUrl` should be called `getFileUrl`. I will leave it to you @dhelma to determine how is the SDK methods naming affecting the general developer sdk usability and methods discoverability.

dghelm:
> I am going to approve it but we should take a look at our SDK methods. If we would want to be consistent with the naming, we already have `getSkylinkUrl` so this new method should probably be `getSkylinkContent`. Otherwise `getSkylinkUrl` should be called `getFileUrl`. I will leave it to you @dhelma to determine how is the SDK methods naming affecting the general developer sdk usability and methods discoverability.

I think File here works, since it doesn't make much sense to use it for Directories. I agree we're going to need a large reconsideration of SDK method names, but, as I understand it from a Discord conversation, we're going to implement that once major features have been shipped so we can take a more complete view at it.

mrcnski (dismissed):

mrcnski (approved):

interstella-5555 (approved):

PR #141, opened 2020-10-26
Revise SkyDB API

mrcnski:


src/crypto.ts
```diff
@@ -17,19 +21,9 @@ export function HashAll(...args: any[]): Uint8Array {
...
-
-// HashFileID hashes the given fileID
-export function HashFileID(fileID: FileID): Uint8Array {
-  return HashAll(
-    encodeNumber(fileID.version),
-    encodeString(fileID.applicationID),
-    encodeNumber(fileID.fileType),
-    encodeString(fileID.filename)
-  );
+// Ashes the given registry entry
```
peterjan: update docstring

src/skydb.ts
```diff
@@ -1,105 +1,65 @@
...
+  const entry = await this.registry.lookup(publicKey, dataKey);
+  if (entry === null) {
     throw new Error("not found");
   }
 
-  // download the data in that Skylink
-  const skylink = existing.value.data;
+  // Download the data in that Skylink
+  // TODO: Replace with download request method.
+  const skylink = entry.value.data;
```
peterjan: should we validate this to be a skylink and throw a meaningful error?

src/skydb.ts
```diff
@@ -1,105 +1,65 @@
...
-  if (!existing) {
+  const entry = await this.registry.lookup(publicKey, dataKey);
+  if (entry === null) {
     throw new Error("not found");
   }
 
-  // download the data in that Skylink
-  const skylink = existing.value.data;
+  // Download the data in that Skylink
+  // TODO: Replace with download request method.
```
peterjan, line 32: remove TODO?

src/skydb.ts
```diff
@@ -1,105 +1,65 @@
...
-
-  // fetch the current value to find out the revision, timeout after 2s
-  let existing: SignedRegistryValue | null;
-  try {
-    existing = await promiseTimeout(this.lookupRegistry(user, fileID), 2000);
-  } catch (error) {
-    existing = null;
+export async function setJSON(this: SkynetClient, privateKey: SecretKey, dataKey: string, json: object, revision: number = -1): Promise<boolean> {
+  // Upload the data to acquire its skylink
+  // TODO: Replace with upload request method.
```
peterjan, line 58: remove TODO

src/skydb.ts
```diff
@@ -1,105 +1,65 @@
...
+  const response = await this.executeRequest({
+    ...this.customOptions,
+    method: "post",
+    endpointPath: "/skynet/skyfile",
+    data: json.toString(),
+  });
+  const skylink = response.skylink;
+
+  const publicKey = pki.ed25519.publicKeyFromPrivateKey({privateKey});
+  if (revision === -1) {
```
peterjan: did we land on this behavior? (works for me)

peterjan: the reason I ask is because I guess the `-1` convention might be considered somewhat of a code smell, even though I don't mind that type of thing

src/crypto.ts
```diff
@@ -1,7 +1,11 @@
 import blake from "blakejs";
-import { RegistryValue } from "./registry";
-import { FileID } from "./skydb";
+import { RegistryEntry } from "./registry";
 import { stringToUint8Array } from "./utils";
+import { pki } from "node-forge";
+
+export type PublicKey = pki.ed25519.NativeBuffer;
```
peterjan, line 6: does this make sense to export? if so, shouldn't we re-export these in index.ts?

src/skydb.ts
```diff
@@ -1,105 +1,65 @@
...
+    // fetch the current value to find out the revision.
+    const entry = await this.registry.lookup(publicKey, dataKey)
+    if (entry) {
+      revision = entry.revision + 1;
+    } else {
+      revision = 0;
+    }
   }
 
   // TODO: we could (/should?) verify here
```
peterjan: we should implement this TODO in the next version imo

src/skydb.ts
mrcnski: I can't yet, the request methods are not available yet, one is in an active PR

src/skydb.ts
mrcnski: Dunno, I got it from your proposal. I think it's fine as a default value but maybe it's idiosyncratic for JS.

src/registry.ts
```diff
@@ -56,8 +43,7 @@ export async function lookupRegistry(
 
   if (response.status === 200) {
     return {
-      value: {
-        tweak: Uint8Array.from(Buffer.from(response.data.tweak)),
+      entry: {
         data: Buffer.from(hexToUint8Array(response.data.data)).toString(),
         revision: parseInt(response.data.revision, 10),
```
interstella-5555, line 64: Is `revision` a string in api?

src/registry.ts
```diff
@@ -56,8 +43,7 @@ export async function lookupRegistry(
 
   if (response.status === 200) {
     return {
-      value: {
-        tweak: Uint8Array.from(Buffer.from(response.data.tweak)),
+      entry: {
```
interstella-5555, line 60: why do we need an `entry` wrapper? couldn't we just return

```javascript
return {
  data: Buffer.from(hexToUint8Array(response.data.data)).toString(),
  revision: parseInt(response.data.revision, 10),
  signature: response.data.signature,
}
```

src/registry.ts
```diff
@@ -67,13 +53,7 @@ export async function lookupRegistry(
   throw new Error(`unexpected response status code ${response.status}`);
```
interstella-5555: You throw on some response codes and some response codes return null. Should we be consistent here or is there any particular reason for that?

src/registry.ts
```diff
@@ -1,52 +1,39 @@
...
+  signature: Signature;
 };
 
-export async function lookupRegistry(
-  this: SkynetClient,
-  user: User,
-  fileID: FileID,
-  customOptions = {}
-): Promise<SignedRegistryValue | null> {
+export async function lookup(this: SkynetClient, publickey: PublicKey, datakey: string, customOptions = {}): Promise<SignedRegistryEntry|null> {
```
interstella-5555: Why is this called `lookup`? It's non standard naming for a getter - would call it a `getRegistryEntry` or `registry.getEntry` if we already namespace with `registry`.

src/registry.ts
```diff
@@ -88,17 +68,12 @@ export async function updateRegistry(
...
-          applicationid: fileID.applicationID,
-          filetype: fileID.fileType,
-          filename: fileID.filename,
-        },
-        revision: srv.value.revision,
-        data: Array.from(Uint8Array.from(Buffer.from(srv.value.data))),
-        signature: Array.from(Uint8Array.from(srv.signature)),
+        datakey,
+        revision: entry.revision,
+        data: Array.from(Uint8Array.from(Buffer.from(entry.data))),
```
interstella-5555: Are you sure these 3 are needed? Wouldn't this suffice? Or even maybe just `Array.from` ? (not tested it, just asking because it looks pretty entangled)
```suggestion
        data: Array.from(Buffer.from(entry.data)),
```

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
-
-// getFile will lookup the entry for given skappID and filename, if it exists it
-// will try and download the file behind the skylink it has found in the entry.
-export async function getFile(this: SkynetClient, user: User, fileID: FileID): Promise<SkyFile> {
+export async function getJSON(this: SkynetClient, publicKey: PublicKey, dataKey: string): Promise<{ json: object, revision: number }|null> {
   // lookup the registry entry
-  const existing = await this.lookupRegistry(user, fileID);
-  if (!existing) {
+  const entry = await this.registry.lookup(publicKey, dataKey);
+  if (entry === null) {
```
interstella-5555, line 27: I would just return `null` on not found, catching errors is just the worst

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
     throw new Error("not found");
   }
 
-  // download the data in that Skylink
-  const skylink = existing.value.data;
+  // Download the data in that Skylink
+  // TODO: Replace with download request method.
+  let skylink: string;
+  try {
+    skylink = parseSkylink(entry.value.data);
```
interstella-5555: `parseSkylink` should probably return `null` on not found as well... not saying to change it here but just sayin'

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
+  } catch (error) {
+    throw new Error(`invalid skylink: ${error}`);
+  }
```
interstella-5555: `parseSkylink` already throws a meaningful error, no need to catch it and rethrow simplified version of it

```suggestion
  const skylink = parseSkylink(entry.value.data);
```

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
     ...this.customOptions,
     method: "get",
     url: this.getSkylinkUrl(skylink),
   });
 
-  // wrap the data in a skyfile and return it
-  const metadata = JSON.parse(response.headers["skynet-file-metadata"]);
-  const file = new SkyFile(new File([response.data], metadata.filename, { type: response.headers["content-type"] }));
-  return file;
+  return { json: response.data, revision: entry.revision };
```
interstella-5555: Let's just call it `data` so it's more generic and we could use the same response type in future `getText` or `getBuffer` methods. We already know it's json because it's the response from `getJson`. Actually, it's not even json - it's json parsed as an object :)

```suggestion
  return { data: response.data, revision: entry.revision };
```

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
+export async function setJSON(this: SkynetClient, privateKey: SecretKey, dataKey: string, json: object, revision: number = -1): Promise<boolean> {
+  // Upload the data to acquire its skylink
+  // TODO: Replace with upload request method.
+  const response = await this.executeRequest({
+    ...this.customOptions,
+    method: "post",
+    endpointPath: "/skynet/skyfile",
+    data: json.toString(),
+  });
+  const skylink = response.skylink;
```
```suggestion
  const { skylink } = await this.executeRequest({
    ...this.customOptions,
    method: "post",
    endpointPath: "/skynet/skyfile",
    data: json.toString(),
  });
```

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
+
+  const publicKey = pki.ed25519.publicKeyFromPrivateKey({privateKey});
+  if (revision === -1) {
```
interstella-5555: I would probably make `revision` an optional argument and check `if (revision) {` here instead.

src/skydb.ts
```diff
@@ -1,105 +1,79 @@
...
-  const signature = user.sign({ message: HashRegistryValue(value) });
+  const signature = pki.ed25519.sign({
+    message: HashRegistryEntry(entry),
+    privateKey,
+  });
 
   // update the registry
-  const updated = await this.updateRegistry(user, fileID, { value, signature });
+  const updated = await this.registry.update(publicKey, dataKey, entry, signature);
   return updated;
```
interstella-5555: No need to await returned value.

```suggestion
  return this.registry.update(publicKey, dataKey, entry, signature);
```

interstella-5555 (changes requested):

src/registry.ts
mrcnski: I don't know, cc @peterjanbrone

mrcnski: @kwypchlo what is more idiomatic, throwing or returning null?

src/registry.ts
mrcnski: It's a number

src/registry.ts
mrcnski: I guess to emphasize that the `signature` is on the entire `entry` object? @peterjanbrone

src/registry.ts
mrcnski: I like `getEntry` but this will require another discussion on Discord, hopefully a short one, but we should get everyone's input.

src/skydb.ts
mrcnski: Wouldn't that be a breaking change? I can create a followup issue.

src/skydb.ts
mrcnski: Looks like you commented on an outdated commit

src/registry.ts
peterjan: Yeah I decided to throw here and return null because this is basically a sanity check to catch developer error.
That or the siad API changed, we basically do not expect to throw this error ever and I wanted to avoid the `try-catch` on the caller side if we want to stay consistent in the expected case.

It's kind of like `build.Critical` or `panic` in siad, at least that was my intention.

src/registry.ts
peterjan: Yeah that's the reason.

src/registry.ts
peterjan: it's a uint64 right so technically we need to handle that, perhaps we should leave a TODO or comment at least?

src/registry.ts
interstella-5555: @peterjanbrone @m-cat SDK should probably abstract that away, otherwise we will end up with people having to check if the response is not null AND also writing a try-catch.

src/skydb.ts
```diff
@@ -37,13 +37,8 @@ export async function setJSON(
...
   // Upload the data to acquire its skylink
   // TODO: Replace with upload request method.
-  const { data } = await this.executeRequest({
-    ...this.customOptions,
-    method: "post",
-    endpointPath: "/skynet/skyfile",
-    data: json.toString(),
-  });
-  const skylink = data.skylink;
+  const file = new File([JSON.stringify(json)], dataKey, { type: "text/plain" });
```
```suggestion
  const file = new File([JSON.stringify(json)], dataKey, { type: "application/json" });
```

src/registry.ts
```diff
@@ -97,5 +97,5 @@ export async function setEntry(
   if (response.status === 204) {
     return true;
   }
-  throw new Error(`unexpected response status code ${response.status}`);
+  return false;
```
interstella-5555: `setEntry` should probably be just. In other case it fails, let's say because revision was wrong, user would not get any error message and would not know how to act - either retry or change the payload, returning `false` doesn't tell user anything useful.

```javascript
return this.executeRequest({
           ...opts,
      	      method: "post",
      	      data: {
      	        publickey: {
                          algorithm: "ed25519",
          	          key: Array.from(publickey),
       	        },
        	        datakey,
        	        revision: entry.revision,
        	        data: Array.from(Buffer.from(entry.data)),
        	        signature: Array.from(signature),
	      },
});
```

interstella-5555 (changes requested):

src/integration.test.ts
```diff
@@ -0,0 +1,31 @@
+import { SkynetClient } from "./client";
```
interstella-5555, line 1: let's move this integration test out of `src` to `e2e` directory

src/crypto.ts
```diff
@@ -17,19 +21,9 @@ export function HashAll(...args: any[]): Uint8Array {
...
-// HashFileID hashes the given fileID
-export function HashFileID(fileID: FileID): Uint8Array {
-  return HashAll(
-    encodeNumber(fileID.version),
-    encodeString(fileID.applicationID),
-    encodeNumber(fileID.fileType),
-    encodeString(fileID.filename)
-  );
+// Hashes the given registry entry.
+export function HashRegistryEntry(registryEntry: RegistryEntry): Uint8Array {
```
peterjan, line 30: this is missing the datakey, you will run into signature verification issues if you do not include it

src/integration.test.ts
```diff
@@ -6,18 +6,17 @@ import { pki } from "node-forge";
 const { publicKey, privateKey } = pki.ed25519.generateKeyPair();
 const client = new SkynetClient("https://siasky.dev");
 
-const appID = "HelloWorld";
+const datakey = "HelloWorld";
 const json = { data: "thisistext" };
 
 // skip - used to verify end-to-end flow
-describe.skip("siasky.dev end to end", () => {
+describe("siasky.dev end to end", () => {
```
interstella-5555: skip that test please, it should be run only manually, otherwise there is a good chance it will fail on our CI and it's not a unit test

src/skydb.ts
```diff
@@ -88,5 +89,5 @@ export async function setJSON(
 
   console.log(9);
   // update the registry
-  return this.registry.setEntry(publicKey, dataKey, entry, signature);
+  this.registry.setEntry(publicKey, dataKey, entry, signature);
```
interstella-5555: if you don't return here, you will need to add `await`

src/skydb.ts
```diff
@@ -49,7 +49,7 @@ export async function setJSON(
   if (!revision) {
     // fetch the current value to find out the revision.
     console.log(3);
-    const entry = await this.registry.getEntry(publicKey, dataKey);
+    const entry = await promiseTimeout(this.registry.getEntry(publicKey, dataKey), 5000);
```
interstella-5555: we shouldn't add an arbitrary timeout here, we should rely on siad either 404 or 200

interstella-5555 (changes requested):
please remove all `console.log` statements left from debugging

src/integration.test.ts
```diff
@@ -0,0 +1,22 @@
+import { SkynetClient } from "./client";
+// import { Buffer } from "buffer";
+import { pki } from "node-forge";
+// import { readData } from "./utils";
```
peterjan: remove unused imports

src/integration.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const { publicKey, privateKey } = pki.ed25519.generateKeyPair();
+const client = new SkynetClient("https://siasky.dev");
+
+const datakey = "HelloWorld";
+const json = { data: "thisistext" };
+
+// skip - used to verify end-to-end flow
+describe.skip("siasky.dev end to end", () => {
+  it("to update the file in the SkyDB", async () => {
```
peterjan: could improve on these descriptions

`describe.skip("skydb integration test")`
`it("should be able to both `setJSON` and `getJSON`")`

src/integration.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    // set the file in the SkyDB
+    await client.db.setJSON(privateKey, datakey, json);
+
+    // get the file in the SkyDB
+    const actual = await client.db.getJSON(publicKey, datakey);
```
peterjan, line 17: Should we be returning the revision number here? I'm not too sure, if the caller cares about it he can just use the registry himself and not go through `setJSON` and `getJSON` which essentially are convenience methods right.

I think `setJSON` and `getJSON` should mirror each other as much as possible and thus only return json.

src/registry.ts
```diff
@@ -1,52 +1,45 @@
...
   customOptions = {}
-): Promise<SignedRegistryValue | null> {
+): Promise<SignedRegistryEntry | null> {
   const opts = {
     ...defaultRegistryOptions,
     ...this.customOptions,
     ...customOptions,
   };
 
+  const userID = publickey.toString("hex");
```
peterjan: (nit) would not call this `userID` and simply inline:
`publickey: ed25519:${publickey.toString("hex")}`

src/registry.ts
```diff
@@ -56,58 +49,47 @@ export async function lookupRegistry(
...
+    publickey: {
+      algorithm: "ed25519",
+      key: Array.from(publickey),
+    },
+    datakey: Buffer.from(HashDataKey(datakey)).toString("hex"),
+    revision: entry.revision,
+    data: Array.from(Buffer.from(entry.data)),
+    signature: Array.from(signature),
+  };
+  console.log(data);
```
peterjan: remove debug code

src/skydb.test.ts
```diff
@@ -1,42 +1,23 @@
...
+const json = { data: "thisistext" };
 
 const portalUrl = defaultSkynetPortalUrl;
 const registryUrl = `${portalUrl}/skynet/registry`;
 const uploadUrl = `${portalUrl}/skynet/skyfile`;
 
 const client = new SkynetClient(portalUrl);
 
-describe.skip("getFile", () => {
+describe("getFile", () => {
```
peterjan: we do not have `getFile` and `setFile` anymore

src/skydb.test.ts
```diff
@@ -47,19 +28,11 @@ describe.skip("getFile", () => {
   it("should perform a lookup and update the window to the skylink url", async () => {
```
peterjan: outdated descriptions

src/skydb.test.ts
```diff
@@ -47,19 +28,11 @@ describe.skip("getFile", () => {
...
-      publickey: `ed25519:${user.id}`,
-      fileid: Buffer.from(
-        JSON.stringify({
-          version: fileID.version,
-          applicationid: fileID.applicationID,
-          filetype: fileID.fileType,
-          filename: fileID.filename,
-        })
-      ).toString("hex"),
+      publickey: `ed25519:${publicKey}`,
```
peterjan: this is confusing I think, in the actual code we are hex encoding this right?

peterjan: maybe we need a helper function for this that also prepends the algorithm

src/skydb.ts
```diff
@@ -1,105 +1,93 @@
...
+export async function getJSON(
+  this: SkynetClient,
+  publicKey: PublicKey,
+  dataKey: string
+): Promise<{ data: Record<string, unknown>; revision: number } | null> {
   // lookup the registry entry
-  const existing = await this.lookupRegistry(user, fileID);
-  if (!existing) {
-    throw new Error("not found");
+  console.log(10);
```
peterjan: remove debug code (commenting once but there's multiple in here)

src/skydb.test.ts
```diff
@@ -1,79 +1,54 @@
...
+describe("getJSON", () => {
   let mock: MockAdapter;
 
   beforeEach(() => {
     mock = new MockAdapter(axios);
     mock.resetHistory();
   });
 
-  it("should perform a lookup and update the window to the skylink url", async () => {
+  // TODO
```
peterjan, line 27: leftover TODO

src/client.ts
```diff
@@ -62,12 +62,16 @@ export class SkynetClient {
   resolveHns = resolveHns;
 
   // SkyDB
-  getFile = getFile;
-  setFile = setFile;
+  db = {
```
peterjan, line 65: not a priority now I guess but this will not be typed properly in the SkynetClient I don't think

peterjan (approved):

src/skydb.test.ts
mrcnski: I set the test to `skip` because it hasn't been fully written yet.

src/skydb.ts
```diff
@@ -1,105 +1,81 @@
...
+    if (entry) {
+      // verify here
+      if (
+        !pki.ed25519.verify({
+          message: HashRegistryEntry(entry.entry),
+          signature: entry.signature,
+          publicKey,
+        })
+      ) {
+        throw new Error("could not verify signature");
```
interstella-5555: Can we make this error a little more user friendly? Give some more details on what could be wrong.

src/client.ts
peterjan: easily fixed by the way, just define an interface 

```
export interface IDB {
  getJSON(
    publicKey: PublicKey,
    dataKey: string
  ): Promise<Record<string, unknown> | null>
  
  setJSON(
    privateKey: SecretKey,
    dataKey: string,
    json: Record<string, unknown>,
    revision?: number
  ): void;
}
```
and do 
```
  resolveHns = resolveHns;

  // SkyDB
  db: IDB = {
    getJSON: getJSON.bind(this),
  ...
```

(rename the interface, it's just for demonstrating purposes)

src/skydb.ts
```diff
@@ -26,9 +24,8 @@ export async function getJSON(
     method: "get",
     url: this.getSkylinkUrl(skylink),
   });
-  console.log(12);
 
-  return { data: response.data, revision: entry.entry.revision };
+  return response.data;
```
interstella-5555: Will we still have access to revision ?

interstella-5555 (changes requested):

src/skydb.ts
mrcnski: Yeah, PJ made the point that the registry functions can be called manually if you need the revision. The SkyDB functions should be easy to use without the user having to keep track of the revision. Makes sense to me IMO.

src/client.ts
mrcnski: Not so easily-fixed, I get typescript errors when I do this. I'm just going to make a follow-up issue for later, thanks.

src/skydb.ts
interstella-5555: Not really, let's say I want to update the content that I just fetched, I will need the revision for that.

src/skydb.ts
mrcnski: https://github.com/NebulousLabs/skynet-js/issues/143

src/client.ts
mrcnski: https://github.com/NebulousLabs/skynet-js/issues/144

src/skydb.ts
mrcnski: You would either use the registry from the beginning or just `setJSON`.

src/skydb.ts
```diff
@@ -1,105 +1,82 @@
...
-export async function getFile(this: SkynetClient, user: User, fileID: FileID): Promise<SkyFile> {
+import { HashRegistryEntry, PublicKey, SecretKey } from "./crypto";
+import { RegistryEntry, SignedRegistryEntry } from "./registry";
+import { parseSkylink, promiseTimeout, trimUriPrefix, uriSkynetPrefix } from "./utils";
+
+export async function getJSON(
+  this: SkynetClient,
+  publicKey: PublicKey,
+  dataKey: string
+): Promise<{ data: Record<string, unknown>; revision: number } | null> {
```
peterjan, line 19: (nit call it `json`?)

src/skydb.ts
```diff
@@ -1,105 +1,82 @@
...
+  // TODO: Replace with upload request method.
+  const file = new File([JSON.stringify(json)], dataKey, { type: "application/json" });
+  const { skylink } = await this.uploadFileRequest(file, this.customOptions);
+
+  const publicKey = pki.ed25519.publicKeyFromPrivateKey({ privateKey });
+  if (!revision) {
+    // fetch the current value to find out the revision.
+    let entry: SignedRegistryEntry;
+    try {
+      entry = await promiseTimeout(this.registry.getEntry(publicKey, dataKey), 5000);
```
peterjan: I think we should go with 30s

src/skydb.ts
```diff
@@ -25,7 +25,7 @@ export async function getJSON(
     url: this.getSkylinkUrl(skylink),
   });
 
-  return { data: response.data, revision: entry.entry.revision };
+  return { json: response.data, revision: entry.entry.revision };
```
interstella-5555: It should be `data` so once we have `getText` it will have the same interface instead of `entry.text`. Why did you change it in the first place?

src/skydb.ts
```diff
@@ -45,7 +45,7 @@ export async function setJSON(
     // fetch the current value to find out the revision.
     let entry: SignedRegistryEntry;
     try {
-      entry = await promiseTimeout(this.registry.getEntry(publicKey, dataKey), 5000);
+      entry = await promiseTimeout(this.registry.getEntry(publicKey, dataKey), 30_000);
```
interstella-5555: it will not stop the request so we might end up with hundreds of pending requests from one app, the timeout should not be in the client

if this is not possible right now, we should use axios `timeout` option, I **think** it will abort the request on timeout but I am not sure

src/skydb.ts
```diff
@@ -2,15 +2,28 @@ import { pki } from "node-forge";
...
 import { RegistryEntry, SignedRegistryEntry } from "./registry";
-import { parseSkylink, promiseTimeout, trimUriPrefix, uriSkynetPrefix } from "./utils";
+import { parseSkylink, trimUriPrefix, uriSkynetPrefix } from "./utils";
 
+/**
+ * Gets the JSON object corresponding to the publicKey and dataKey.
+ * @param publicKey - The user public key.
+ * @param dataKey - The key of the data to fetch for the given user.
+ * @param [customOptions={}] - Additional settings that can optionally be set.
+ * @param [customOptions.timeout_seconds=30] - Timeout in seconds for the registry lookup.
```
interstella-5555: `timeout_seconds` should be `timeout` and accept milliseconds - it's the standard in javascript to use milliseconds (check setTimeout as an example)

interstella-5555 (changes requested):

interstella-5555 (approved):

PR #146, opened 2020-10-28
Add keyPairFromSeed

mrcnski:


src/client.ts
```diff
@@ -63,6 +63,7 @@ export class SkynetClient {
 
   // SkyDB
   db = {
+    keyPairFromSeed: keyPairFromSeed.bind(this),
```
peterjan: I would not define this on the db.

interstella-5555: agreed, it's also something registry will make use of

src/skydb.ts
```diff
@@ -66,7 +66,13 @@ export async function setJSON(
     try {
       entry = await this.registry.getEntry(publicKey, dataKey, opts);
 
-      // verify here
+      revision = entry.entry.revision + 1;
+    } catch (err) {
+      revision = 0;
+    }
+
+    if (entry) {
```
interstella-5555: flatten those two ifs into one if

src/utils.ts
```diff
@@ -55,6 +57,15 @@ export function getRootDirectory(file: File): string {
   return path.normalize(dir).slice(root.length).split(path.sep)[0];
 }
 
+/**
+ * Generates a public and private key from a provided, secure seed.
+ * @param seed - A secure seed.
+ */
+export function keyPairFromSeed(seed: string): { publicKey: PublicKey; privateKey: SecretKey } {
+  seed = pkcs5.pbkdf2(seed, "", 1000, 32, md.sha256.create());
```
interstella-5555, line 66: Can we add a comment on why is this needed here ? @peterjanbrone maybe you could collaborate on that

src/integration.test.ts
```diff
@@ -1,7 +1,6 @@
-import { SkynetClient } from "./client";
-import { pki } from "node-forge";
+import { keyPairFromSeed, SkynetClient } from "./index";
 
-const { publicKey, privateKey } = pki.ed25519.generateKeyPair();
+const { publicKey, privateKey } = keyPairFromSeed("test");
```
interstella-5555: this should have a randomly generated seed

src/utils.ts
```diff
@@ -55,6 +57,15 @@ export function getRootDirectory(file: File): string {
   return path.normalize(dir).slice(root.length).split(path.sep)[0];
 }
 
+/**
+ * Generates a public and private key from a provided, secure seed.
+ * @param seed - A secure seed.
+ */
+export function keyPairFromSeed(seed: string): { publicKey: PublicKey; privateKey: SecretKey } {
```
interstella-5555, line 64: I would prefix that function name with a verb like `generate` or `create`

interstella-5555 (changes requested):

src/utils.ts
mrcnski: I think it's fine as-is.

src/integration.test.ts
```diff
@@ -1,6 +1,7 @@
-import { keyPairFromSeed, SkynetClient } from "./index";
+import { SkynetClient } from "./client";
+import { pki } from "node-forge";
 
-const { publicKey, privateKey } = keyPairFromSeed("test");
+const { publicKey, privateKey } = pki.ed25519.generateKeyPair();
```
interstella-5555: this is not what I meant, you should still use `keyPairFromSeed` but the seed you pass should be random instead of `"test"`

interstella-5555 (changes requested):

interstella-5555 (approved):

PR #152, opened 2020-11-01
Add crypto seed API

mrcnski:
Crypto seed API for SkyID:

+ create master key from seed (a known/pre-existing seed is provided and we turn it into a key)
+ create new random master key (we will generate randomness and return both a key and also the seed for that key)
+ derive child key from master key (input is the master seed plus a string)

### DIFF
--- a/src/crypto.ts
+++ b/src/crypto.ts
@@ -2,7 +2,7 @@ import { pki, pkcs5, md } from "node-forge";
 import blake from "blakejs";
 import { RegistryEntry } from "./registry";
 import { stringToUint8Array } from "./utils";
-import { randomBytes } from "randombytes";
+import randomBytes from "randombytes";
 
 export type PublicKey = pki.ed25519.NativeBuffer;
 export type SecretKey = pki.ed25519.NativeBuffer;
--- a/src/crypto.ts
+++ b/src/crypto.ts
@@ -2,26 +2,7 @@ import { pki, pkcs5, md } from "node-forge";
 import blake from "blakejs";
 import { RegistryEntry } from "./registry";
 import { stringToUint8Array } from "./utils";
-let crypto: any;
-
-const isBrowser =
-  typeof window !== "undefined" &&
-  Object.prototype.hasOwnProperty.call(window, "Window") &&
-  window instanceof window.Window;
-
-function hasStrongRandom() {
-  return "crypto" in window && window["crypto"] !== null;
-}
-
-if (isBrowser) {
-  if (hasStrongRandom()) {
-    crypto = window["crypto"];
-  } else {
-    throw "Error, you browser does not support secure random generation.";
-  }
-} else {
-  crypto = require("crypto");
-}
+import { randomBytes } from "randombytes";
 
 export type PublicKey = pki.ed25519.NativeBuffer;
 export type SecretKey = pki.ed25519.NativeBuffer;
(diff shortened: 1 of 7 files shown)


DaWe35:
Hi!
Thank you for bringing it together on Sunday. Anyway, I have some feedback, so I'll write some comments to the code.

src/crypto.ts
```diff
@@ -53,3 +53,33 @@ function encodeString(str: string): Uint8Array {
...
 }
+
+export function deriveChildSeed(masterSeed: string, seed: string): string {
+  return HashAll(masterSeed, seed).toString();
+}
+
+export function generateKeyPairAndSeed(length?: 64): { publicKey: PublicKey; privateKey: SecretKey; seed: string } {
+  const seed = makeSeed(length);
+  const { publicKey, privateKey } = keyPairFromSeed(seed);
+  return { publicKey, privateKey, seed };
```
```suggestion
  return { ...keyPairFromSeed(seed), seed };
```

src/crypto.ts
```diff
@@ -53,3 +53,33 @@ function encodeString(str: string): Uint8Array {
...
+
+function makeSeed(length: number) {
+  let result = "";
+  const characters = "abcdefghijklmnopqrstuvwxyz";
+  const charactersLength = characters.length;
+  for (let i = 0; i < length; i++) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+  return result;
+}
```
interstella-5555: That function seems weak for a seed generator 
- there is a chance that `Math.floor(Math.random() * charactersLength)` returns an index out of bounds if `Math.random()` returns `1`
- `Math.random()` is not a cryptographically safe random function - there is a number of articles on the internet about it

We should use a safer algorithm for a random seed generation.

interstella-5555 (changes requested):

src/crypto.ts
DaWe35: For SkyID I used this, maybe it helps: https://github.com/escada-finance/sia-js/blob/afbdb7eb8edbb050c2ac7272d243b02fd5604a25/lib/KeyPair.js#L23
It is because I need to generate a mnemonic.

mrcnski: Thanks @DaWe35, I just saw your comment. I implemented it only for browser and looks like it fails when running the tests in Node. Will nick your function 😉

src/crypto.ts
```diff
@@ -53,3 +73,38 @@ function encodeString(str: string): Uint8Array {
...
+function makeSeed(length: number): string {
+  if (isBrowser) {
+    const array = new Uint8Array(length);
+    // Use built-in cryptographically-secure random number generator.
+    window.crypto.getRandomValues(array);
+    return Buffer.from(array).toString("hex");
+  } else {
+    const array = crypto.randomBytes(length);
+    return Buffer.from(array).toString("hex");
+  }
```
interstella-5555: You won't need all these checks if you'd use https://www.npmjs.com/package/randombytes

interstella-5555 (changes requested):

DaWe35 (changes requested):
I think you forgot to add 
```
,
  "browser": {
		"crypto": false
	}
```
to package.json (after devDependencies).
Webpack throws an error without: 
kép
https://stackoverflow.com/questions/54162297/module-not-found-error-cant-resolve-crypto

src/crypto.ts
DaWe35: > You won't need all these checks if you'd use https://www.npmjs.com/package/randombytes

Browsers and Node also have built-in crypto module, I think it is better than using an additional dependency. Browsers also have built-in window.crypto

interstella-5555: from https://www.npmjs.com/package/randombytes

> In node you just get crypto.randomBytes, but in the browser it uses .crypto/msCrypto.getRandomValues

DaWe35 (approved):

interstella-5555 (approved):

PR #166, opened 2020-11-05
Clean up SkyDB API

### LINKED ISSUE #161: `dataKey` is passed twice into `skynetClient.registry.setEntry(...)`
**`dataKey` is passed twice into `skynetClient.registry.setEntry(...)`**

example from https://siasky.net/docs/#setting-data-on-the-registry
```javascript
import { SkynetClient, keyPairFromSeed } from "skynet-js";

const client = new SkynetClient();
const { publicKey, privateKey } = keyPairFromSeed("this seed should be fairly long for security");

const dataKey = "foo";
const data = "bar";
const revision = 0;
const entry = { datakey, data, revision };

async function setEntryExample() {
  try {
    await client.registry.setEntry(privateKey, dataKey, entry);
  } catch (error) {
    console.log(error);
  }
}
```

Variable `dataKey` is passed once to create `entry`
```javascript
const entry = { datakey, data, revision };
```
and then also as a second parameter to `setEntry`
```javascript
client.registry.setEntry(privateKey, dataKey, entry);
```

**Solution proposals**
```javascript
// (my choice)
const entry = { datakey, data, revision };
client.registry.setEntry(privateKey, entry);
```
or
```javascript
const entry = { data, revision };
client.registry.setEntry(privateKey, dataKey, entry);
```
or
```javascript
client.registry.setEntry(privateKey, dataKey, data, revision);
```

### LINKED ISSUE #162: Change request: keyPairFromSeed and generateKeyPairAndSeed should return keys as strings
Functions `keyPairFromSeed` and `generateKeyPairAndSeed` should return keys as strings, not buffers, otherwise if I want to print the publickey to user, I need to install a buffer package and use toString("hex") myself which is an extra step and I think totally unnecessary.

closes https://github.com/NebulousLabs/skynet-js/issues/165

### LINKED ISSUE #163: Change request: `skynetClient.registry.getEntry` should return `{ entry: null }` on entry not found
Function `skynetClient.registry.getEntry` now returns `null` if entry is not found so we need to write 
```javascript
const entry = skynetClient.registry.getEntry(...);

if(entry) {
  console.log(entry.entry.data);
}
``` 
and it works fine (not even mentioning weird `entry.entry` chain) but I think usually people would write 
```javascript
const { entry } = skynetClient.registry.getEntry(...);
``` 
but it will throw an error if `getEntry` returns `null`. 

I think we should return `{ entry: null }` if the entry is not found instead of just `null` so it's easier for people to destructure the result and saves us weird accessing `entry.entry`.

### LINKED ISSUE #164: Some functions have first letter capitalised
functions like `HashAll`, `HashDataKey`, `NewHash`, `HashRegistryEntry`  should have first letter lowercase

### LINKED ISSUE #165: Change request: `keyPairFromSeed` should include `specifier` in returned value
Function `keyPairFromSeed` should also include `specifier` in the response object, set to `"ed25519"` (it wouldn't be needed if private key and public key would be strings in the response so close this one if https://github.com/NebulousLabs/skynet-js/issues/162 is accepted).

### LINKED ISSUE #167: Inconsistency in method naming
`generateKeyPairAndSeed` and `keyPairFromSeed`

One starts with a verb `generate`, the other doesn't. Since we're doing some breaking changes anyway, let's change `keyPairFromSeed` to `generateKeyPairFromSeed`.

mrcnski:
Closes #161
Closes #162
Closes #163
Closes #164
Closes #165

Closes #167


src/crypto.ts
```diff
@@ -8,35 +8,35 @@ export type PublicKey = pki.ed25519.NativeBuffer;
...
-// HashAll takes all given arguments and hashes them.
-export function HashAll(...args: any[]): Uint8Array {
-  const h = NewHash();
+// Takes all given arguments and hashes them.
+export function hashAll(...args: any[]): Uint8Array {
+  const h = newHash();
   for (let i = 0; i < args.length; i++) {
     blake.blake2bUpdate(h, args[i]);
   }
   return blake.blake2bFinal(h);
```
interstella-5555: - we can remove `newHash` function completely and replace it here with the implementation - it's just used here once, not exported and the blake2 implementation is not being abstracted so there is no need to create a separate function for it
- try to use meaningful variable names `h` -> `hash`

```suggestion
  const hash = blake.blake2bInit(32, null);
  for (let i = 0; i < args.length; i++) {
    blake.blake2bUpdate(hash, args[i]);
  }
  return blake.blake2bFinal(hash);
```

src/crypto.ts
```diff
@@ -8,35 +8,35 @@ export type PublicKey = pki.ed25519.NativeBuffer;
 export type SecretKey = pki.ed25519.NativeBuffer;
 export type Signature = pki.ed25519.NativeBuffer;
 
-// NewHash returns a blake2b 256bit hasher.
-function NewHash() {
+// Returns a blake2b 256bit hasher.
+function newHash() {
   return blake.blake2bInit(32, null);
 }
```
```suggestion
```

interstella-5555 (commented):
Great PR! Only cosmetic changes requested, otherwise peachy.

Added two more small improvements that I missed originally. I hope we could include them both in this PR.

https://github.com/NebulousLabs/skynet-js/issues/168
https://github.com/NebulousLabs/skynet-js/issues/167

src/crypto.ts
mrcnski: This is basically a copy/paste from Sia. I'll keep `newHash` for now since I may use it when implementing encryption.

interstella-5555 (approved):
I don't really like `gen` prefix instead of `generate` since it just degrades the readability for no apparent reason but that's not something I will hold this pull request for.

PR #176, opened 2020-11-09
Add getEntryUrl

### LINKED ISSUE #168: new method request: getRegistryEntryUrl
I need to print a url after creating the registry entry. Currently I do it like this:

```javascript
const encodedPublicKey = encodeURIComponent(`ed25519:${publicKey}`);
const encodedDataKey = encodeURIComponent(Buffer.from(HashDataKey(dataKey)).toString("hex"));

console.log(`https://siasky.net/skynet/registry?publickey=${encodedPublicKey}&datakey=${encodedDataKey}`);
```

I would like a function called `getRegistryEntryUrl` that would allow me to modify the above code to be
```javascript
console.log(skynetClient.registry.getEntryUrl(publicKey, dataKey));
```

mrcnski:
Closes #168

### DIFF
--- a/src/registry.test.ts
+++ b/src/registry.test.ts
@@ -1,0 +1,17 @@
+import { SkynetClient, defaultSkynetPortalUrl, genKeyPairAndSeed } from "./index";
+import { hashDataKey } from "./crypto";
+
+const portalUrl = defaultSkynetPortalUrl;
+const client = new SkynetClient(portalUrl);
+const { publicKey } = genKeyPairAndSeed();
+const dataKey = "app";
+
+describe("getEntryUrl", () => {
+  it("should generate the correct registry url for the given entry", () => {
+    const url = client.registry.getEntryUrl(publicKey, dataKey);
+    const encodedPK = encodeURIComponent(`ed25519:${publicKey}`);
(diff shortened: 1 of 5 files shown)


src/registry.test.ts
```diff
@@ -0,0 +1,17 @@
...
+const portalUrl = defaultSkynetPortalUrl;
+const client = new SkynetClient(portalUrl);
+const { publicKey } = genKeyPairAndSeed();
+const dataKey = "app";
+
+describe("getEntryUrl", () => {
+  it("should generate the correct registry url for the given entry", () => {
+    const url = client.registry.getEntryUrl(publicKey, dataKey);
+    const encodedPK = encodeURIComponent(`ed25519:${publicKey}`);
+    const encodedDK = encodeURIComponent(Buffer.from(hashDataKey(dataKey)).toString("hex"));
```
interstella-5555: I would actually hardcode those as strings because if something changes in Buffer in next node or something, we will not catch an issue.

interstella-5555 (changes requested):
looks great, just one suggestion for improvement in unit test

interstella-5555 (approved):

PR #177, opened 2020-11-09
add getRegistry, setRegistry functions (do NOT merge)

DaWe35:
I just wanted to update a registry to point to an already uploaded skyfile. I think this commit can be helpful, but it is not tested because skynet-js completely broken for me.

### DIFF
--- a/src/skydb-direct.ts
+++ b/src/skydb-direct.ts
@@ -1,0 +1,73 @@
+import { pki } from "node-forge";
+import { SkynetClient } from "./client";
+import { RegistryEntry, SignedRegistryEntry } from "./registry";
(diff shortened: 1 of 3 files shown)


interstella-5555:
why not use registry.setEntry and registry.getEntry ?

src/skydb-direct.ts
```diff
@@ -0,0 +1,73 @@
...
+    ...this.customOptions,
+    ...customOptions,
+  };
+
+  const privateKeyBuffer = Buffer.from(privateKey, "hex");
+
+  // Upload the data to acquire its skylink
+  // TODO: Replace with upload request method.
+
+  if (!revision) {
```
DaWe35, line 51: @kwypchlo because I need this if statement and was easier to copy & delete some code

PR #180, opened 2020-11-10
Make deriveChildSeed return a hex string and add toHexString

### LINKED ISSUE #178: genKeyPairFromSeed() returns bytearray instead of HEX
genKeyPairFromSeed() returns something weird bytearray instead of HEX

Solution: Use [this function](https://stackoverflow.com/questions/34309988/byte-array-to-hex-string-conversion-in-javascript) instead of .toString('hex')

mrcnski:
Fixes #178


src/utils.ts
```diff
@@ -108,18 +108,30 @@ export function randomNumber(low: number, high: number): number {
...
   return new Uint8Array(str.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
 }
 
-// readData is a helper function that uses a FileReader to read the contents of
-// the given file
+/**
+ * Convert a byte array to a hex string.
+ * From https://stackoverflow.com/a/44608819.
+ */
+export function toHexString(byteArray: Uint8Array) {
```
```suggestion
export function toHexString(byteArray: Uint8Array): string {
```

interstella-5555 (changes requested):
one small improvement, otherwise 👍

DaWe35:
Hi, setFile function broken, fixed here: https://github.com/NebulousLabs/skynet-js/pull/181

src/skydb.ts
```diff
@@ -61,13 +61,12 @@ export async function setJSON(
   const file = new File([JSON.stringify(json)], dataKey, { type: "application/json" });
   const { skylink } = await this.uploadFileRequest(file, opts);
 
-  if (!revision) {
+  if (revision == -1) {
```
```suggestion
  if (revision === undefined) {
```

src/skydb.ts
```diff
@@ -46,7 +46,7 @@ export async function setJSON(
   privateKey: string,
   dataKey: string,
   json: Record<string, unknown>,
-  revision?: number,
+  revision = -1,
```
```suggestion
  revision?: number,
```

interstella-5555 (changes requested):

src/utils.ts
```diff
@@ -108,18 +108,30 @@ export function randomNumber(low: number, high: number): number {
...
+/**
+ * Convert a byte array to a hex string.
+ * From https://stackoverflow.com/a/44608819.
+ */
+export function toHexString(byteArray: Uint8Array): string {
+  let s = "";
+  byteArray.forEach(function (byte) {
+    s += ("0" + (byte & 0xff).toString(16)).slice(-2);
+  });
+  return s;
```
ilkamo, line 130:
```suggestion
  return Buffer.from(byteArray).toString('hex');
```
A fancy way to do the same with Buffer.

mrcnski: Thanks. This is what we had before but @kwypchlo wants to move away from the `buffer` package (which is basically a polyfill since `Buffer` is not available everywhere).

ro-tex (approved):
LGTM

PR #195, opened 2020-11-18
Coverage

mrcnski:


mrcnski:
We went from coverage in the 80s to 98% coverage.

There are still about 3 uncovered lines that I need Karol's help with writing tests for.

mrcnski:
@kwypchlo Need some help with getRootDirectory, getRelativeFilePath, the test cases I wrote don't seem to make sense. Not sure if we can even test them in node.

package.json
```diff
@@ -21,6 +21,17 @@
...
+      "**/*.ts"
+    ],
+    "coverageThreshold": {
+      "global": {
+        "branches": 98,
+        "functions": 98,
+        "lines": 98,
+        "statements": 98
+      }
+    },
```
ro-tex: Setting this to the exact percentage where we are at the moment means that if you change a couple of lines in an uncovered method it will prevent you from pushing. If we set it to 90% that would be just as good, as we won't be able to make meaningful code changes without adding tests but we'll be able to quickly fix something if need be.

Also, while I'm usually in favour of some high code coverage percentage (typically 80%), I would strongly advocate against going for 100% coverage. There are two main reasons:
1. The last couple of percent are usually unreasonably difficult to obtain - rarely encountered edge case errors which are hard to simulate and take unreasonable amount of time to write tests for.
2. People tend to strive for those 100% and once they reach them they get a false sense of security - "my code is 100% tested, everything is fine". This is very dangerous because while every line of your code is being covered, not every execution path is and it's very easy to forget very important edge cases while having your code 100% covered. This is a little fallacy that be fall victims of because we misinterpret those 100%. 

In order to protect against those two (mostly the second, the first is just a little annoying but not dangerous), I would suggest to intentionally not try to achieve 100% coverage.

src/client.ts
```diff
@@ -15,25 +15,53 @@ import {
...
   portalUrl: string;
   customOptions: CustomClientOptions;
 
   /**
-   * The Skynet Client which can be used to access Skynet.
+   * @class
+   *
    * @param [portalUrl] The portal URL to use to access Skynet, if specified. To use the default portal while passing custom options, use ""
-   * @param [customOptions] Configuration for the client
+   * @param [customOptions] Configuration for the client.
```
ro-tex: PP: Can we keep the field definitions above the methods? Maybe it's just me but I find it very confusing to have these mixed.

src/utils.ts
```diff
@@ -193,23 +300,32 @@ export function trimUriPrefix(str: string, prefix: string): string {
...
+ * Converts a hex encoded string to a uint8 array
+ *
+ * @param str - The string to convert.
+ * @returns - The uint8 array.
+ */
 export function hexToUint8Array(str: string): Uint8Array {
   return new Uint8Array(str.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
 }
 
+// From https://stackoverflow.com/a/44608819.
```
ro-tex: nit: This could be a `@see`.

src/utils.ts
```diff
@@ -137,15 +208,23 @@ export function parseSkylink(skylinkStr: string, opts: any = {}): string {
...
-function parseSkylinkBase32(skylinkStr: string, opts: any = {}): string {
+/**
+ * Parses the given string for a base32 skylink.
+ *
+ * @param skylinkStr - Base32 skylink.
+ * @param [opts] - Additional settings that can optionally be set.
+ * @returns - The base32 skylink.
+ */
+export function parseSkylinkBase32(skylinkStr: string, opts?: ParseSkylinkBase32Options): string | null {
+  opts = { ...opts };
```
ro-tex: Does this basically ensure that `opts` is not null?

src/utils.ts
```diff
@@ -70,6 +131,9 @@ export function getRelativeFilePath(file: File): string {
   return path.join(...relative, base);
 }
 
+/**
+ * @param file
+ */
```
ro-tex, line 164: We seem to need more text here - a description and a return type. I'm not sure why it's also complaining about the `file`.

src/utils.ts
```diff
@@ -62,6 +120,9 @@ function getFilePath(
   return file.webkitRelativePath || file.path || file.name;
 }
 
+/**
+ * @param file
+ */
```
ro-tex, line 150: Same as https://github.com/NebulousLabs/skynet-js/pull/195#discussion_r528602639.

src/utils.test.ts
```diff
@@ -39,6 +42,34 @@ describe("convertSkylinkToBase32", () => {
...
+    expect(getRelativeFilePath(file)).toEqual(directory);
+  });
+});
+
+describe("getRootDirectory", () => {
+  // TODO: Is this right?
+  const filepaths = [
+    ["abc/def", "."],
+    ["./abc/def", "."],
+    ["/abc/def", "."],
```
ro-tex: This looks wrong. I don't understand why it passes. The third case makes no sense to me.

src/download.ts
```diff
@@ -125,40 +179,60 @@ export function getHnsUrl(this: SkynetClient, domain: string, customOptions: any
...
   domain = trimUriPrefix(domain, uriHandshakeResolverPrefix);
   return makeUrl(this.portalUrl, opts.endpointPath, domain);
 }
 
-export async function getMetadata(this: SkynetClient, skylink: string, customOptions: any = {}) {
+/**
+ * Gets only the metadata for the given skylink without the contents.
+ *
+ * @param this - SkynetClient
+ * @param skylinkStr - Skylink string. See `downloadFile`.
```
ro-tex: Are you sure about this naming? I consistently see `skylink` everywhere else. Is this value somehow different?

ro-tex (commented):
LGTM overall.

The main thing I'd like to see changed is the expectation about code coverage levels. That might need a proper discussion, so we can agree to it as a team.

src/utils.ts
mrcnski: Right. I could have given `opts` a default value of `{}` instead, but I wanted to be consistent with how we fill in the options in methods like `uploadFile` and `downloadFile`. Maybe a default value is more clear though. I think I'll change it since `opts = { ...opts }` does look kind of weird.

src/utils.ts
mrcnski: Agreed, waiting on Karol for help with these functions because I don't remember how they work. I'd like to do that in a followup though so we don't hold up this PR.

src/utils.test.ts
mrcnski: I have the same feeling. I think these functions do not work properly in node. I would like to fix this in a followup with Karol's help.

src/download.ts
mrcnski: I want to distinguish somehow between a skylink and a string that could contain a skylink (e.g. a portal URL). Not sure what to name it, but I don't think `skylink` here was a great name because we still have to extract the skylink from this string.

src/utils.ts
mrcnski: Tanks, didn't know about that!

package.json
mrcnski: > The last couple of percent are usually unreasonably difficult to obtain - rarely encountered edge case errors which are hard to simulate and take unreasonable amount of time to write tests for.

Yeah, there are a couple of cases I haven't been able to write tests for for this reason. For example, one line that only is run in the browser and never triggered in tests.

A workaround would be to (sparingly) add tags to disable coverage reporting for specific lines. The above case might be the only one where it's absolutely needed. Considering that, I believe we can attain and maintain 100% coverage.

> People tend to strive for those 100% and once they reach them they get a false sense of security

Interesting psychological point. I feel like having all code covered is better than having some statements/branches not tested at all. It's true that 100% does not really mean 100% and there's a lot of combinations that aren't tested (true full coverage not being realistically possible.) We just have to keep that in mind when writing tests.

Do you have any sources I could read up further on regarding this point? It reminds me of the "helmets make you feel safer and thus more likely to get in an accident" argument. It would be nice if someone has collected actual data on bugs caught/missed with/without full coverage.

> If we set it to 90% that would be just as good, as we won't be able to make meaningful code changes without adding tests but we'll be able to quickly fix something if need be.

Hmm I'm not sure we want to be pushing quick untested changes anyway. Having coverage reporting, and strict rules about every line being tested, earlier would have prevented a few bugs from making it to production.

ro-tex: > "helmets make you feel safer and thus more likely to get in an accident"

Ha ha, yeah, it is something like that!

> Do you have any sources I could read up further on regarding this point? 

No, I'm afraid I don't - I might try digging around though. My only source is my experience but that's always biased, based on the people I worked with in the past.

Last point: setting the minimum coverage level to 90% does not stop you in any way from adding all the tests you want and reaching 100%.

Anyway, this was only a suggestion. If you feel 100% is appropriate and reasonably achievable then let's go for that.

src/download.ts
ro-tex: I see. That makes sense. Maybe `skylinkUrl` can be more descriptive? It's just that `skylinkStr` did not immediately tell me that it's called that because there is some additional fluff in the string around the skylink itself.

ro-tex (dismissed):
Let's wait for Karol's review but LGTM.

.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
 node_modules
 dist
-yarn-error.log
\ No newline at end of file
+yarn-error.log
+src/coverage
```
interstella-5555: please generate coverage outside of src directory - just `coverage` is fine

package.json
```diff
@@ -21,6 +21,17 @@
   "jest": {
     "testTimeout": 60000,
     "clearMocks": true,
+    "collectCoverageFrom": [
+      "**/*.ts"
+    ],
```
interstella-5555, line 26:
```suggestion
    "collectCoverageFrom": [
      "**/*.ts",
      "!**/node_modules/**"
    ],
```

src/crypto.ts
```diff
@@ -47,38 +91,60 @@ function encodeNumber(num: number): Uint8Array {
...
   // Get a 32-byte seed.
   seed = pkcs5.pbkdf2(seed, "", 1000, 32, md.sha256.create());
   const { publicKey, privateKey } = pki.ed25519.generateKeyPair({ seed });
   return { publicKey: toHexString(publicKey), privateKey: toHexString(privateKey) };
 }
 
+/**
+ * Generates a random seed of the given length in bytes.
+ *
+ * @param length - Length of th eseed in bytes.
```
```suggestion
 * @param length - Length of the seed in bytes.
```

src/skydb.test.ts
```diff
@@ -79,6 +75,30 @@ describe("setJSON", () => {
...
+    // mock a successful upload
+    mock.onPost(uploadUrl).reply(200, { skylink });
+
+    // mock a successful registry update
+    mock.onPost(registryUrl).reply(204);
+
+    // set data
+    const updated = await client.db.setJSON(privateKey, dataKey, json, revision + 1);
+
+    expect(updated);
```
interstella-5555, line 95: I think you forgot an assertion here

src/utils.test.ts
```diff
@@ -39,6 +42,34 @@ describe("convertSkylinkToBase32", () => {
...
+  ];
+
+  it.each(filepaths)("the root directory for a file with the path %s should be %s", (filepath, directory) => {
+    const file = new File(["test contents"], filepath);
```
interstella-5555:
```suggestion
    const file = new File(["test contents"], "foo");
    file.path = filepath;
```

You need to spoof the `file.path` prop because if you provide the file path as a name like you did, it will escape all `/` into `:` and it will not detect the root dir properly.

src/utils.test.ts
interstella-5555: responded below

src/utils.ts
```diff
@@ -5,40 +5,87 @@ import path from "path-browserify";
...
+export type BaseCustomOptions = CustomClientOptions & {
+  endpointPath?: string;
+};
+
+/**
+ * Parse skylink options.
+ *
+ * @property [fromSubdomain] - Whether to parse the skylink as a base32 subdomain in a URL.
+ * @property [includePath] - Whether to include the path after the skylink.
```
interstella-5555: add example to the comment like you did below please

src/utils.ts
```diff
@@ -154,32 +229,60 @@ function parseSkylinkBase32(skylinkStr: string, opts: any = {}): string {
   const matchHostname = parsed.hostname.match(SKYLINK_SUBDOMAIN_REGEX);
   if (matchHostname) {
     if (opts.onlyPath) {
-      return parsed.pathname;
+      const path = trimSuffix(parsed.pathname, "/");
+      return path;
```
```suggestion
      return trimSuffix(parsed.pathname, "/");
```

src/utils.ts
```diff
@@ -219,27 +331,20 @@ export function toHexString(byteArray: Uint8Array): string {
...
+ * @param file - The file.
+ * @returns - The mime type.
  */
 export function getFileMimeType(file: File): string {
   if (file.type) return file.type;
-  const extension = file.name.slice(file.name.lastIndexOf(".") + 1);
-  if (extension) {
+  let { ext } = path.parse(file.name);
+  ext = trimPrefix(ext, ".");
+  if (ext != "") {
```
```suggestion
  if (ext !== "") {
```

interstella-5555 (changes requested):

src/utils.test.ts
ro-tex: @kwypchlo I see two mentions of "responded below" but I can't find the response itself. Could you please add a link to it?

src/skydb.test.ts
mrcnski: I actually fixed this in the bigint PR. That one's going to get merged first

package.json
mrcnski: We shouldn't need this since `rootDir` is `src`

src/download.ts
mrcnski: Sounds good to me.

interstella-5555 (dismissed):

src/skydb.test.ts
```diff
@@ -88,7 +88,7 @@ describe("setJSON", () => {
     mock.onPost(registryUrl).reply(204);
 
     // set data
-    const updated = await client.db.setJSON(privateKey, dataKey, json, revision + 1);
+    const updated = await client.db.setJSON(privateKey, dataKey, json, BigInt(revision + 1));
```
interstella-5555, line 93: revision should be mocked as BigInt on line 22

src/utils.ts
```diff
@@ -80,7 +80,7 @@ export function addUrlQuery(url: string, query: Record<string, unknown>): string
  * @throws - Will throw if the int does not fit in 64 bits.
  * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN | MDN Demo}
  */
-export function checkUint64(int: bigint) {
+export function checkUint64(int: bigint): void {
```
interstella-5555: in javascript (at least) we call this type of type guarding functions assertions, can you rename it to `assertUint64` ?

interstella-5555 (changes requested):

interstella-5555 (dismissed):

src/client.ts
```diff
@@ -76,11 +94,26 @@ export class SkynetClient {
...
   /**
    * Creates and executes a request.
-   * @param {Object} config - Configuration for the request. See docs for constructor for the full list of options.
+   *
+   * @param config - Configuration for the request.
+   * @returns - The response from axios.
+   * @throws - Will throw if unimplemented options have been passed in.
    */
-  protected executeRequest(config: any): Promise<AxiosResponse> {
+  protected executeRequest(config: RequestConfig): Promise<AxiosResponse> {
```
ro-tex, line 116: It's so nice watching the `any`s going away. :)

ro-tex (dismissed):
LGTM

peterjan (approved):

ro-tex (approved):

PR #207, opened 2020-11-24
Bigint revision

mrcnski:
The revision number needs to be a `bigint` instead of `number` to support the full range of accepted revisions on the Sia side (uint64 has a wider range than allowed in 53 bits, the max for `number` in JS).

**This is a breaking change.**

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -78,7 +78,7 @@ export class SkynetClient {
    * Creates and executes a request.
    * @param {Object} config - Configuration for the request. See docs for constructor for the full list of options.
    */
-  executeRequest(config: any): Promise<AxiosResponse> {
+  protected executeRequest(config: any): Promise<AxiosResponse> {
     if (config.skykeyName || config.skykeyId) {
       throw new Error("Unimplemented: skykeys have not been implemented in this SDK");
     }
@@ -91,7 +91,10 @@ export class SkynetClient {
       url = addUrlQuery(url, config.query);
     }
 
-    const headers = { ...config.headers, "User-Agent": config.customUserAgent && config.customUserAgent };
+    const headers = { ...config.headers };
+    if (config.customUserAgent) {
+      headers["User-Agent"] = config.customUserAgent;
+    }
 
     return axios({
       url,
--- a/src/client.ts
+++ b/src/client.ts
@@ -91,8 +91,8 @@ export class SkynetClient {
       url = addUrlQuery(url, config.query);
     }
 
-    // No other headers.
-    const headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
+    let headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
+    headers = { ...config.headers, ...headers };
 
     return axios({
       url,
@@ -108,6 +108,8 @@ export class SkynetClient {
           config.onUploadProgress(progress, event);
         },
       timeout: config.timeout,
+      transformRequest: config.transformRequest,
+      transformResponse: config.transformResponse,
 
       maxContentLength: Infinity,
       maxBodyLength: Infinity,
--- a/src/client.ts
+++ b/src/client.ts
@@ -91,8 +91,7 @@ export class SkynetClient {
       url = addUrlQuery(url, config.query);
     }
 
-    let headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
-    headers = { ...config.headers, ...headers };
+    const headers = { ...config.headers, "User-Agent": config.customUserAgent && config.customUserAgent };
 
     return axios({
       url,
--- a/src/crypto.test.ts
+++ b/src/crypto.test.ts
@@ -43,7 +43,7 @@ describe("hashRegistryValue", () => {
     const hash = hashRegistryEntry({
       datakey: "HelloWorld",
       data: "abc",
-      revision: 123456789,
+      revision: BigInt(123456789),
     });
 
     expect(toHexString(hash)).toEqual(h);
--- a/src/crypto.test.ts
+++ b/src/crypto.test.ts
@@ -1,5 +1,5 @@
 import { deriveChildSeed, encodeBigintAsUint64, genKeyPairFromSeed, hashRegistryEntry } from "./crypto";
-import { maxint, toHexString } from "./utils";
+import { MAX_REVISION, toHexString } from "./utils";
 
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
@@ -59,10 +59,10 @@ describe("encodeBigint", () => {
     expect(encodeBigintAsUint64(BigInt(255))).toEqualUint8Array(new Uint8Array([255, 0, 0, 0, 0, 0, 0, 0]));
     expect(encodeBigintAsUint64(BigInt(256))).toEqualUint8Array(new Uint8Array([0, 1, 0, 0, 0, 0, 0, 0]));
     expect(encodeBigintAsUint64(BigInt(256))).not.toEqualUint8Array(new Uint8Array([1, 1, 0, 0, 0, 0, 0, 0]));
-    expect(encodeBigintAsUint64(BigInt(maxint))).toEqualUint8Array(
+    expect(encodeBigintAsUint64(MAX_REVISION)).toEqualUint8Array(
       new Uint8Array([255, 255, 255, 255, 255, 255, 255, 255])
     );
-    expect(() => encodeBigintAsUint64(BigInt(maxint) + BigInt(1))).toThrowError(
+    expect(() => encodeBigintAsUint64(MAX_REVISION + BigInt(1))).toThrowError(
       "Argument 18446744073709551616 does not fit in a 64-bit unsigned integer; exceeds 2^64-1"
     );
   });
--- a/src/crypto.test.ts
+++ b/src/crypto.test.ts
@@ -1,12 +1,36 @@
 import { deriveChildSeed, encodeBigintAsUint64, genKeyPairFromSeed, hashRegistryEntry } from "./crypto";
-import { areEqualUint8Arrays, toHexString } from "./utils";
+import { toHexString } from "./utils";
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toEqualUint8Array(argument: Uint8Array): R;
+    }
+  }
+}
+
+expect.extend({
+  // source https://stackoverflow.com/a/60818105/6085242
+  toEqualUint8Array(received: Uint8Array, argument: Uint8Array) {
+    if (received.length !== argument.length) {
+      return { pass: false, message: () => `expected ${received} to equal ${argument}` };
+    }
+    for (let i = 0; i < received.length; i++) {
+      if (received[i] !== argument[i]) {
+        return { pass: false, message: () => `expected ${received} to equal ${argument}` };
+      }
+    }
+    return { pass: true, message: () => `expected ${received} not to equal ${argument}` };
+  },
+});
 
 describe("areEqualUint8Arrays", () => {
   it("should correctly check whether uint8arrays are equal", () => {
-    expect(areEqualUint8Arrays(new Uint8Array([0]), new Uint8Array([0]))).toBeTruthy();
-    expect(areEqualUint8Arrays(new Uint8Array([1, 1, 0]), new Uint8Array([1, 1, 0]))).toBeTruthy();
-    expect(areEqualUint8Arrays(new Uint8Array([1, 0, 0]), new Uint8Array([1, 1, 0]))).toBeFalsy();
-    expect(areEqualUint8Arrays(new Uint8Array([1, 1, 0]), new Uint8Array([1, 1, 0, 0]))).toBeFalsy();
+    expect(new Uint8Array([0])).toEqualUint8Array(new Uint8Array([0]));
+    expect(new Uint8Array([1, 1, 0])).toEqualUint8Array(new Uint8Array([1, 1, 0]));
+    expect(new Uint8Array([1, 0, 0])).not.toEqualUint8Array(new Uint8Array([1, 1, 0]));
+    expect(new Uint8Array([1, 1, 0])).not.toEqualUint8Array(new Uint8Array([1, 1, 0, 0]));
   });
 });
 
@@ -33,22 +57,13 @@ describe("encodeBigint", () => {
   it("should correctly encode bigints", () => {
     const maxint = "18446744073709551615";
 
-    expect(areEqualUint8Arrays(encodeBigintAsUint64(BigInt(0)), new Uint8Array([0, 0, 0, 0, 0, 0, 0, 0]))).toBeTruthy();
-    expect(
-      areEqualUint8Arrays(encodeBigintAsUint64(BigInt(255)), new Uint8Array([255, 0, 0, 0, 0, 0, 0, 0]))
-    ).toBeTruthy();
-    expect(
-      areEqualUint8Arrays(encodeBigintAsUint64(BigInt(256)), new Uint8Array([0, 1, 0, 0, 0, 0, 0, 0]))
-    ).toBeTruthy();
-    expect(
-      areEqualUint8Arrays(encodeBigintAsUint64(BigInt(256)), new Uint8Array([1, 1, 0, 0, 0, 0, 0, 0]))
-    ).toBeFalsy();
-    expect(
-      areEqualUint8Arrays(
-        encodeBigintAsUint64(BigInt(maxint)),
-        new Uint8Array([255, 255, 255, 255, 255, 255, 255, 255])
-      )
-    ).toBeTruthy();
+    expect(encodeBigintAsUint64(BigInt(0))).toEqualUint8Array(new Uint8Array([0, 0, 0, 0, 0, 0, 0, 0]));
+    expect(encodeBigintAsUint64(BigInt(255))).toEqualUint8Array(new Uint8Array([255, 0, 0, 0, 0, 0, 0, 0]));
+    expect(encodeBigintAsUint64(BigInt(256))).toEqualUint8Array(new Uint8Array([0, 1, 0, 0, 0, 0, 0, 0]));
+    expect(encodeBigintAsUint64(BigInt(256))).not.toEqualUint8Array(new Uint8Array([1, 1, 0, 0, 0, 0, 0, 0]));
+    expect(encodeBigintAsUint64(BigInt(maxint))).toEqualUint8Array(
+      new Uint8Array([255, 255, 255, 255, 255, 255, 255, 255])
+    );
     expect(() => encodeBigintAsUint64(BigInt(maxint) + BigInt(1))).toThrowError("Received int > 2^64-1");
   });
 });
--- a/src/crypto.test.ts
+++ b/src/crypto.test.ts
@@ -64,7 +64,9 @@ describe("encodeBigint", () => {
     expect(encodeBigintAsUint64(BigInt(maxint))).toEqualUint8Array(
       new Uint8Array([255, 255, 255, 255, 255, 255, 255, 255])
     );
-    expect(() => encodeBigintAsUint64(BigInt(maxint) + BigInt(1))).toThrowError("Received int > 2^64-1");
+    expect(() => encodeBigintAsUint64(BigInt(maxint) + BigInt(1))).toThrowError(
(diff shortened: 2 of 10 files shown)


src/skydb.test.ts
```diff
@@ -107,6 +107,6 @@ describe("setJSON", () => {
 
     const data = JSON.parse(mock.history.post[1].data);
     expect(data).toBeDefined();
-    expect(data.revision).toEqual(0);
+    expect(data.revision).toEqual("0");
```
interstella-5555: it shouldn't be a string though right?

src/registry.ts
```diff
@@ -143,5 +156,14 @@ export async function setEntry(
...
     method: "post",
     data,
+    // Transform the request to remove quotes, since the revision needs to be
+    // parsed as a uint64 on the Go side.
+    transformRequest: function (data: unknown) {
+      // Convert the object data to JSON.
+      let json = JSON.stringify(data);
+      // Change the revision value from a string to a JSON integer.
+      json = json.replace(/"revision":"([0-9]+)"/, '"revision":$1');
+      return json;
```
interstella-5555:
```suggestion
      const json = JSON.stringify(data);
      // Change the revision value from a string to a JSON integer.
      return json.replace(/"revision":"([0-9]+)"/, '"revision":$1');
```

src/registry.ts
```diff
@@ -56,6 +56,15 @@ export async function getEntry(
...
       },
       timeout: opts.timeout,
+      // Transform the response to add quotes, since uint64 cannot be accurately
+      // read by JS so the revision needs to be parsed as a string.
+      transformResponse: function (data: string) {
+        // Change the revision value from a JSON integer to a string.
+        data = data.replace(/"revision":([0-9]+)/, '"revision":"$1"');
+        // Convert the JSON data to an object.
+        data = JSON.parse(data);
+        return data;
```
interstella-5555:
```suggestion
        return JSON.parse(data);
```

src/crypto.ts
```diff
@@ -47,6 +47,22 @@ function encodeNumber(num: number): Uint8Array {
...
+ */
+function encodeBigInt(int: bigint): Uint8Array {
+  const encoded = new Uint8Array(8);
+  for (let index = 0; index < encoded.length; index++) {
+    const byte = int & BigInt(0xff);
+    encoded[index] = Number(byte);
+    int = int >> BigInt(8);
+  }
+  return encoded;
+}
```
interstella-5555: we should unit test this function

src/integration.test.ts
```diff
@@ -3,18 +3,38 @@ import { genKeyPairAndSeed, SkynetClient } from "./index";
...
+    // Set the file in the SkyDB.
+    await client.db.setJSON(privateKey, dataKey, json);
 
     // get the file in the SkyDB
-    const actual = await client.db.getJSON(publicKey, datakey);
+    let actual = await client.db.getJSON(publicKey, dataKey);
+    expect(actual.data).toEqual(json);
+    expect(actual.revision).toEqual(BigInt(0));
+
+    // Set the revision to the max allowed.
```
interstella-5555: that test is in fact 3 separate tests - please extract them out

interstella-5555 (changes requested):

src/skydb.test.ts
mrcnski: You commented on an old commit.

src/crypto.ts
```diff
@@ -47,6 +47,28 @@ function encodeNumber(num: number): Uint8Array {
...
+/**
+ * Converts the given bigint into a uint8 array.
+ *
+ * @param int - Bigint to encode.
+ * @returns - Bigint encoded as a byte array.
+ */
+export function encodeBigintAsUint64(int: bigint): Uint8Array {
+  // Assert the input is 64 bits.
+  const newint = BigInt.asUintN(64, int);
+  if (newint != int) {
```
```suggestion
  if (newint !== int) {
```

src/crypto.ts
```diff
@@ -47,6 +47,28 @@ function encodeNumber(num: number): Uint8Array {
...
+    throw new Error("Received int > 2^64-1");
```
interstella-5555: Add a current argument value to the error message, it might help debugging what value exceeded the limit.
```suggestion
    throw new Error(`Argument ${int} exceeds 2^64-1`);
```

src/registry.ts
```diff
@@ -113,6 +121,12 @@ export async function setEntry(
   entry: RegistryEntry,
   customOptions = {}
 ): Promise<void> {
+  // Assert the input is 64 bits.
+  const newint = BigInt.asUintN(64, entry.revision);
+  if (newint != entry.revision) {
```
interstella-5555:
```suggestion
  if (newint !== entry.revision) {
```

src/skydb.ts
```diff
@@ -70,9 +70,15 @@ export async function setJSON(
...
+      revision = entry.entry.revision + BigInt(1);
     } catch (err) {
-      revision = 0;
+      revision = BigInt(0);
+    }
+  } else {
+    // Assert the input is 64 bits.
+    const newint = BigInt.asUintN(64, revision);
+    if (newint != revision) {
+      throw new Error("Received revision number > 2^64-1");
```
interstella-5555: This is potentially dev facing error - we should denote that the 2^64-1 is the hard limit of 64 bit int.

src/skydb.ts
```diff
@@ -70,9 +70,15 @@ export async function setJSON(
...
-      revision = entry.entry.revision + 1;
+      revision = entry.entry.revision + BigInt(1);
     } catch (err) {
-      revision = 0;
+      revision = BigInt(0);
+    }
+  } else {
+    // Assert the input is 64 bits.
+    const newint = BigInt.asUintN(64, revision);
+    if (newint != revision) {
```
```suggestion
    if (newint !== revision) {
```

interstella-5555 (changes requested):
I added a commit that changed the uint8array comparison for tests. The way it was used, in case a test failed it would only tell "false does not equal true" or similar.  Using expect.extend we leverage built in comparison system and verbose failure error messages.

Some minor changes left, otherwise good but I'm not really super happy about passing revision as BigInt. I'll raise a small discussion on discord.

src/client.ts
```diff
@@ -91,8 +91,8 @@ export class SkynetClient {
       url = addUrlQuery(url, config.query);
     }
 
-    // No other headers.
-    const headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
+    let headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
```
peterjan: This is a weird way of writing it. If `config.customUserAgent` is not defined or falsy, headers will initially be undefined, and you're destructing it in the next line. It works though, but I think this is better:

```
const headers = config.customUserAgent
    ? { ...config.headers, "User-Agent": config.customUserAgent} 
    : config.headers
```

src/registry.ts
```diff
@@ -56,6 +56,14 @@ export async function getEntry(
         datakey: toHexString(hashDataKey(dataKey)),
       },
       timeout: opts.timeout,
+      // Transform the response to add quotes, since uint64 cannot be accurately
+      // read by JS so the revision needs to be parsed as a string.
+      transformResponse: function (data: string) {
```
peterjan, line 67: Hm, well obviously a bit dirty at first sight, but probably there are not many alternative ways of solving this. I'm assuming you spent time googling this right? I do think though that the `.replace` with the mention on `revision` as a field is something we can improve upon.

Why don't we for example use this library: https://www.npmjs.com/package/json-bigint
There are many more like this, might be worth investigating those as well.
This way we could have the SDK handle all future bigints coming from the API in the same way.
By default it uses `bignumber.js` but they have an option `useNativeBigInt` that would work for us I think.

src/registry.ts
```diff
@@ -113,6 +121,12 @@ export async function setEntry(
   entry: RegistryEntry,
   customOptions = {}
 ): Promise<void> {
+  // Assert the input is 64 bits.
+  const newint = BigInt.asUintN(64, entry.revision);
```
peterjan: I think this is fine but I would use the function as defined in the `BigInt.asUintN` documentation page literally: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN 
That way I think it's more readable and we can cover it with a small unit test.

Side-note: I also kind of like the 'n' notion, like `10n`, as opposed to the explicit `BigInt(10)` type constructor

src/integration.test.ts
```diff
@@ -3,18 +3,43 @@ import { genKeyPairAndSeed, SkynetClient } from "./index";
...
-const json = { data: "thisistext" };
+const dataKey = "HelloWorld";
+const maxint = BigInt("18446744073709551615"); // max uint64
 
 // skip - used to verify end-to-end flow
 describe.skip("siasky.dev end to end integration test", () => {
-  it("should be able to both setJSON and getJSON", async () => {
-    // set the file in the SkyDB
-    await client.db.setJSON(privateKey, datakey, json);
+  // NOTE: These tests must run sequentially.
```
peterjan: Why though? The tests are not dependant on each other right? 

If you move `const { publicKey, privateKey } = genKeyPairAndSeed();` inside of the test itself wouldn't you be able to run them in parallel? I don't think it's really an issue but I also don't see why we need to overcomplicate it and reference how you can run tests sequentially if they don't need to be run that way necessarily.

src/registry.ts
```diff
@@ -143,5 +159,13 @@ export async function setEntry(
     ...opts,
     method: "post",
     data,
+    // Transform the request to remove quotes, since the revision needs to be
+    // parsed as a uint64 on the Go side.
+    transformRequest: function (data: unknown) {
```
peterjan, line 170: To be safe we probably need to do some type checking on `data` here. In the current implementation I guess it's not necessary, because we're applying it only to a subset of the requests, but if we were to use the `json-bigint` library it would become generic and get called on all requests and responses and we'd have to verify `data` is an object here and then call `JSONbig.stringify(data)` on it.

I only glanced at the documentation though, but there's bound to be quite some articles on how to do it. Here are a couple I found:
https://medium.com/@itstalwar15/handling-bigint-in-axios-using-json-big-in-javascript-d915ae85ffc0
https://stackoverflow.com/questions/43787712/axios-how-to-deal-with-big-integers
https://github.com/axios/axios/issues/983

peterjan (changes requested):
One final question perhaps, couple of weeks ago we discussed briefly `bigint` libraries and back then we weren't able to use the native BigInt for reasons that I can't remember clearly. Something to do with the encryption PR you were working on at the time.

Do we still have those concerns w/regards to the native`bigint`library? What changed? 
Will we be needing two bigint implementations soon when we ad encryption?

src/registry.ts
mrcnski: Thanks, I'll make that change.

I like `10n` as well but it is only available in ECMA 2020

mrcnski:
> Do we still have those concerns w/regards to the nativebigintlibrary? What changed?
Will we be needing two bigint implementations soon when we ad encryption?

@peterjanbrone We couldn't use native bigints because we needed a bigint method that wasn't supported in a couple of major browsers. I used `JSBI` for convenience because it would be easy to replace that with native bigints once that method was available.

Now I think we should use native bigints. They are part of the public API here, so users of the SDK wouldnt need to import `jsbi`. The method I needed for encryption can just be written manually, if it is still needed in the future.

src/registry.ts
interstella-5555: I actually checked that library earlier and it's going to pull bignumber.js into the package regardless whether we use it or not so if it's just one use case then I think we should be fine with a replace hack.

mrcnski: It seems a bit hacky but this is precisely the intended use of the transform functions. I don't think it's a huge deal.

src/skydb.ts
mrcnski: No longer applies, changing this

src/crypto.ts
mrcnski: No longer applies, changing this

src/client.ts
mrcnski: Rewrote it a bit differently so it's easier to extend in the future

mrcnski: For reference:

```javascript
const headers = { ...config.headers, "User-Agent": config.customUserAgent && config.customUserAgent };
```

src/integration.test.ts
mrcnski: Sure

src/registry.ts
mrcnski: That looks pretty elegant. If we ever have more uint64s in the API we can switch over. For now I'd prefer not to introduce another dependency.

In the present code I don't think type checking is required here

src/utils.ts
```diff
@@ -33,6 +33,25 @@ export function addUrlQuery(url: string, query: Record<string, unknown>): string
...
 
+/**
+ * Checks if the provided bigint can fit in a 64-bit unsigned integer.
+ *
+ * @param int - The provided integer.
+ * @throws - Will throw if the int does not fit in 64 bits.
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN | MDN Demo}
+ */
+export function checkUint64(int: bigint) {
+  const maxint = BigInt("18446744073709551615"); // max uint64
```
interstella-5555: this could be defined on top of the file, otherwise we build the big integer each time this function is called - minor but low hanging fruit perf optimisation

src/client.ts
```diff
@@ -91,8 +91,7 @@ export class SkynetClient {
       url = addUrlQuery(url, config.query);
     }
 
-    // No other headers.
-    const headers = config.customUserAgent && { "User-Agent": config.customUserAgent };
+    const headers = { ...config.headers, "User-Agent": config.customUserAgent && config.customUserAgent };
```
peterjan: still think this is dirty - it  yieds:
```
{
foo:"bar",
User-Agent:undefined
}
```
if `config.customUserAgent` is undefined.

peterjan (dismissed):

src/client.ts
interstella-5555:
```suggestion
    const headers = { ...config.headers, "User-Agent": config.customUserAgent };
```

src/client.ts
```diff
@@ -108,6 +107,8 @@ export class SkynetClient {
           config.onUploadProgress(progress, event);
         },
       timeout: config.timeout,
+      transformRequest: config.transformRequest,
+      transformResponse: config.transformResponse,
```
interstella-5555, line 114: this is not something devs will be able to override right? we don't want to expose that to devs

interstella-5555 (changes requested):

src/client.ts
peterjan: Hm yeah.. I don't really mind (at all actually) - but this is still dirty. You'll end up with  `User-Agent` being undefined. But present in the headers. So `Object.keys()` on `headers` will list it..

I don't see the downside of what I proposed at all.
If we ever add one, then you rewrite like this:

```
let headers = config.headers
if (config.customUserAgent) {
   headers[ "User-Agent"] = config.customUserAgent
}
if (config.otherHeader) {
   headers[ "Other-Header"] = config.otherHeader
}
```

src/client.ts
mrcnski: Since we agree, I'll make `executeRequest` protected as we are breaking compatibility anyway.

The main use of that for devs would have been to get file contents (not sure if it was actually used that way). We have that coming soon in Dan's PR.

interstella-5555 (dismissed):

src/skydb.ts
```diff
@@ -70,10 +70,13 @@ export async function setJSON(
...
       entry = await this.registry.getEntry(toHexString(publicKey), dataKey, opts);
-      revision = entry.entry.revision + 1;
+      revision = entry.entry.revision + BigInt(1);
     } catch (err) {
-      revision = 0;
+      revision = BigInt(0);
     }
+  } else {
+    // Assert the input is 64 bits.
+    checkUint64(revision);
```
ro-tex, line 81: Shouldn't we verify that the revision number is valid, i.e. `old_rev+1`? Or are we doing that somewhere else?

src/registry.ts
```diff
@@ -56,6 +56,14 @@ export async function getEntry(
         datakey: toHexString(hashDataKey(dataKey)),
       },
       timeout: opts.timeout,
+      // Transform the response to add quotes, since uint64 cannot be accurately
+      // read by JS so the revision needs to be parsed as a string.
+      transformResponse: function (data: string) {
+        // Change the revision value from a JSON integer to a string.
+        data = data.replace(/"revision":([0-9]+)/, '"revision":"$1"');
```
ro-tex: This regex should allow for some whitespace between the colon and the number. It doesn't need to preserve it, though.

src/registry.ts
```diff
@@ -143,5 +156,13 @@ export async function setEntry(
     ...opts,
     method: "post",
     data,
+    // Transform the request to remove quotes, since the revision needs to be
+    // parsed as a uint64 on the Go side.
+    transformRequest: function (data: unknown) {
+      // Convert the object data to JSON.
+      const json = JSON.stringify(data);
+      // Change the revision value from a string to a JSON integer.
+      return json.replace(/"revision":"([0-9]+)"/, '"revision":$1');
```
ro-tex: We have this regex in more than one spot. This is a possible place to introduce a bug. Let's move it to a const and use that instead.

src/crypto.test.ts
```diff
@@ -20,6 +53,23 @@ describe("deriveChildSeed", () => {
   });
 });
 
+describe("encodeBigint", () => {
+  it("should correctly encode bigints", () => {
+    const maxint = "18446744073709551615";
```
ro-tex: We define `maxint` in another place (or two?) as well. Let's move it to a global const.

src/integration.test.ts
```diff
@@ -1,20 +1,44 @@
...
-const { publicKey, privateKey } = genKeyPairAndSeed();
 const client = new SkynetClient("https://siasky.dev");
 
-const datakey = "HelloWorld";
-const json = { data: "thisistext" };
+const dataKey = "HelloWorld";
+const maxint = BigInt("18446744073709551615"); // max uint64
 
 // skip - used to verify end-to-end flow
 describe.skip("siasky.dev end to end integration test", () => {
```
ro-tex: Are we ready to enable this suite?

src/crypto.ts
```diff
@@ -47,6 +47,25 @@ function encodeNumber(num: number): Uint8Array {
   return encoded;
 }
 
+/**
+ * Converts the given bigint into a uint8 array.
+ *
+ * @param int - Bigint to encode.
+ * @returns - Bigint encoded as a byte array.
```
ro-tex, line 54: Can we add `@throws` that explains that it might throw if the int is too big? I wouldn't expect that, so it would be helpful.

ro-tex (changes requested):
LGTM, overall. A couple of small improvement requested.

src/skydb.ts
mrcnski: If it's provided, then we use it in the call to Sia. If it's not valid, Sia will report an error. Otherwise we would have to do an extra API call just to get the old revision

src/registry.ts
mrcnski: It's actually a different regex, one has quotes around the revision value and one doesn't

src/skydb.ts
ro-tex: That makes sense.

src/registry.ts
ro-tex: Ah, right, this one basically does the reverse. OK but I would still feel better if those are in named constants. Especially if they are so visually similar.

mrcnski: Just saw this comment. Okay, will do.

src/registry.ts
```diff
@@ -14,6 +14,12 @@ const defaultSetEntryOptions = {
   ...defaultOptions("/skynet/registry"),
 };
 
+// Regex for JSON revision value without quotes.
+const regexRevisionNoQuotes = /"revision":\s?([0-9]+)/;
+
+// Regex for JSON revision value with quotes.
+const regexRevisionWithQuotes = /"revision":\s?"([0-9]+)"/;
```
ro-tex: The `?` matches only one space.
```suggestion
// Regex for JSON revision value without quotes.
const regexRevisionNoQuotes = /"revision":\s*([0-9]+)/;

// Regex for JSON revision value with quotes.
const regexRevisionWithQuotes = /"revision":\s*"([0-9]+)"/;
```

peterjan (dismissed):

ro-tex (dismissed):

interstella-5555 (approved):

ro-tex (approved):

PR #208, opened 2020-11-27
split node version and build

interstella-5555:


tobowers:
Adding a few notes from my endeavours here.

I think the codebase itself is pretty easy to cleanup... for download I did this:

```typescript
export function internalDownload(this: SkynetClient, url: string, customOptions?: CustomDownloadOptions) {

  if (isBrowser) {
    window.location.assign(url);
  } else {
    if (customOptions?.receiver) {
      this.executeRequest({
        ...opts,
        method: "get",
        url,
      }).then((resp) => {
        if (!customOptions.receiver) {
          throw new Error("unexpected error, no receiver");
        }
        return customOptions.receiver(resp.data);
      });
    }
  }
}
```

Where I allow the customDownloadOptions to specify a function to receive the bytes.

A few place to clean up `File` by allowing a `Buffer` also. 

Most of the work is in the tests. I would highly recommend *not* separating the test files, but instead have the files run once with default and then again with `--env node`. Splitting the files will cause maintenance head aches. Most of the work is in cleaning up those mocks and spys... but it's not crazy bad - I bet it'd only take a day... basically you'd create a function to do the right thing for mocking so like `mockDownload(url)` which would do axios or window.

Per convo in discord on 28 Jan 2021... leave uploadFile to take File or Buffer but in node add `uploadPath` which would read from the fs.

mrcnski:
@kwypchlo Since conditional builds are seeming like too much work I think we don't have to do that. Just mark in the docs the functions that are browser-only and node-only, and raise a runtime error if a user tries to call e.g. `uploadPath` (per comment above) from the browser.

mrcnski:
@quorumcontrol released this which may be helpful to you @kwypchlo :

https://github.com/quorumcontrol/skydb

mrcnski:
Closing in flavor of #338, most of this PR was incorporated there.

PR #216, opened 2020-12-02
Version update to v2.10.0

mrcnski:


interstella-5555 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## [2.10.0]
+
+_Note: this version contains breaking changes: the registry entry `revision` is now a `bigint` instead of `number`, and the `executeRequest` method on the client is no longer available._
+
```
ro-tex, line 6: This means that it shouldn't be version 2.10.0 but 3.0.0.

mrcnski: This isn't the first time we had breaking changes in v2 😛

ro-tex (changes requested):
If we're breaking compatibility we need to increment the major version. Unless we've specifically said that we are not following semver.

CHANGELOG.md
ro-tex: So... no semver? Can we at least announce that somewhere prominent?

I mean, people's `npm` and `yarn` will upgrade this package for them and break their applications, this is not ideal, especially if the fix is a version increment.

mrcnski:
We'll be releasing v3 instead of another minor version. I'll get a couple more breaking changes in before then.

PR #218, opened 2020-12-03
[BREAKING] Pass registry timeout as API parameter

mrcnski:
## Breaking changes

- We no longer accept `timeout` for any API methods apart from `getEntry`. It was never explicitly supported for other methods, but was possible, and it was passed along to axios.
- The `timeout` option is now in seconds rather than ms


src/integration.test.ts
```diff
@@ -43,3 +43,15 @@ describe("siasky.dev end to end integration test", () => {
...
+describe("Registry end to end integration tests", () => {
+  it("Should timeout after the given timeout time in ms", async () => {
+    const { publicKey } = genKeyPairAndSeed();
+
+    // Try getting an inexistant entry.
+    const { entry, signature } = await client.registry.getEntry(publicKey, "foo");
+
+    expect(entry).toBeNull();
+    expect(signature).toBeNull();
+  });
```
interstella-5555: you don't pass the timeout anywhere

interstella-5555 (changes requested):

src/integration.test.ts
```diff
@@ -43,3 +43,15 @@ describe("siasky.dev end to end integration test", () => {
...
+});
```
ro-tex: Can we add a test where we pass a `timeout` to the `set` method and see that it doesn't affect it?

src/integration.test.ts
mrcnski: I wasn't able to test the timeout really, even for a nonexistant entry the portal got back to me quite fast. I wanted to keep that test but forgot to update the description

I added a test that uses a timeout that is too large and gets rejected, this proves that it gets passed to Sia.

src/integration.test.ts
```diff
@@ -54,4 +55,29 @@ describe("Registry end to end integration tests", () => {
...
+
+    expect(returnedEntry).toEqual(entry);
+
+    // Try getting an entry with an excessive timeout.
+    const { entry: returnedEntry2, signature } = await client.registry.getEntry(publicKey, dataKey, {
+      timeout: MAX_GET_ENTRY_TIMEOUT + 1,
+    });
+
+    expect(returnedEntry2).toBeNull();
+    expect(signature).toBeNull();
```
interstella-5555: you said this would throw an error ?

mrcnski: > Sia will return an error and getEntry will return destructured nulls :)

Currently any error from Sia gets caught and we return nulls, nothing changed there.

interstella-5555: Ok I will create a follow up for that.

src/integration.test.ts
```diff
@@ -54,4 +55,29 @@ describe("Registry end to end integration tests", () => {
...
+  });
```
interstella-5555: there are 2 separate tests here

interstella-5555 (changes requested):

src/integration.test.ts
```diff
@@ -43,3 +44,40 @@ describe("siasky.dev end to end integration test", () => {
...
+    const { publicKey, privateKey } = genKeyPairAndSeed();
+
+    const entry = {
+      datakey: dataKey,
+      data: "foo",
+      revision: BigInt(0),
+    };
+
+    // Try passing minimum timeout, should have no effect in setEntry.
+    await client.registry.setEntry(privateKey, entry, { timeout: 1 });
```
ro-tex: I am not sure how this verifies that the setting doesn't have any effect on `setEntry`. Do we assume that we cannot make a `set` registry call in 1 second? That shouldn't be the case.

src/registry.ts
```diff
@@ -5,15 +5,22 @@ import { addUrlQuery, checkUint64, defaultOptions, hexToUint8Array, makeUrl, toH
...
   ...defaultOptions("/skynet/registry"),
-  timeout: 5_000,
+  timeout: 5,
 };
 
 const defaultSetEntryOptions = {
   ...defaultOptions("/skynet/registry"),
 };
 
+export const MAX_GET_ENTRY_TIMEOUT = 300; // 5 minutes
```
peterjan, line 39: In `siad` we valuate whether the timeout is between 1 and 300s, we might want to do the same here.

src/integration.test.ts
mrcnski: If you mean the two getEntry calls, the first one confirms that the entry has been set before we try to make an invalid call.

mrcnski: Ended up refactoring this test. It no longer needed to be an integration test because we error on invalid timeout on the client side.

ro-tex (dismissed):

interstella-5555 (dismissed):

src/registry.ts
```diff
@@ -92,6 +94,10 @@ export async function getEntry(
     ...customOptions,
   };
 
+  if (opts.timeout > MAX_GET_ENTRY_TIMEOUT || opts.timeout < 1) {
+    throw new Error(`Invalid 'timeout' parameter, needs to be between 1s and ${MAX_GET_ENTRY_TIMEOUT}s`);
+  }
```
interstella-5555: does it accept floats too?

interstella-5555 (approved):

ro-tex (approved):

PR #221, opened 2020-12-04
Enable strictNullChecks

mrcnski:


ro-tex (dismissed):

dghelm (dismissed):

dghelm (dismissed):

src/download.ts
```diff
@@ -136,7 +136,7 @@ export function getSkylinkUrl(this: SkynetClient, skylinkUrl: string, customOpti
   let url;
   if (opts.subdomain) {
     // Get the path from the skylink.
-    const skylinkPath = parseSkylink(skylinkUrl, { onlyPath: true });
+    const skylinkPath = parseSkylink(skylinkUrl, { onlyPath: true }) ?? "";
```
ro-tex, line 154: Would it make sense to build this directly into `parseSkylink`? I mean the defaulting to an empty string instead of a `nil`/`undefined`.

mrcnski: `null` seems to be a standard way to indicate "not found", for example this is how `String.prototype.match()` works

utils/testing.ts
```diff
@@ -36,6 +36,9 @@ export async function compareFormData(formData: Record<string, unknown>, entries
     // Read the file asynchronously.
     const reader = new FileReader();
     reader.onload = function (e) {
+      if (e.target === null) {
+        throw new Error("e.target === null");
+      }
```
interstella-5555: Why do you need that in the test? Would an error here just break the function, not call the "expect" below and mark that test case false-positive ?

utils/testing.ts
```diff
@@ -36,9 +36,8 @@ export async function compareFormData(formData: Record<string, unknown>, entries
     // Read the file asynchronously.
     const reader = new FileReader();
     reader.onload = function (e) {
-      if (e.target === null) {
-        throw new Error("e.target === null");
-      }
+      expect(e.target).not.toBeNull();
+      if (e.target === null) return; // Make the null-check happy.
```
interstella-5555: I assume typescript complains on `e.target.result` that `e.target` can be null? I would just drop this `if` and replace that expect below with `expect(e.target?.result).toEqual(expectedData);`

```javascript
> e = {}
{}
> e.target.result
Uncaught TypeError: Cannot read property 'result' of undefined
> e.target?.result
undefined
```

interstella-5555 (changes requested):

dghelm (dismissed):

src/download.ts
interstella-5555: This `parseSkylink(skylinkUrl, { onlyPath: true }) ?? ""` will not work because `??` only assigns the value on the right if the value on the left is `undefined` while `parseSkylink` never returns `undefined`.

You should move `const skylinkPath = ...` below the `if (skylink === null) {` check and you will not need that conditional `""` assignment because it won't be null then.

mrcnski: MDN:

> The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.

I think it's correct as written, I can't do your suggestion because the null check is for a different call to parseSkylink

interstella-5555: Oh you're right then, sorry. Anyway, you should be able to move that code below that check and you won't need to check for nulls.

mrcnski: We can't do that or the strict-null-check lint will complain

ro-tex: Just FYI, this is resolved on my end.

ro-tex (dismissed):

interstella-5555 (dismissed):

PR #225, opened 2020-12-07
Don't prepend key prefix if it's present already

mrcnski:


src/registry.test.ts
```diff
@@ -56,13 +56,20 @@ describe("getEntry", () => {
...
-    const publicKey = "c1197e1275fbf570d21dde01a00af83ed4a743d1884e4a09cebce0dd21ae254c";
-    const encodedPK = "ed25519%3Ac1197e1275fbf570d21dde01a00af83ed4a743d1884e4a09cebce0dd21ae254c";
-    const encodedDK = "7c96a0537ab2aaac9cfe0eca217732f4e10791625b4ab4c17e4d91c8078713b9";
     const url = client.registry.getEntryUrl(publicKey, dataKey);
 
     expect(url).toEqual(`${portalUrl}/skynet/registry?publickey=${encodedPK}&datakey=${encodedDK}`);
   });
+
+  it("should trim the prefix if it is provided", () => {
+    const url = client.registry.getEntryUrl(`ed25519:${publicKey}`, dataKey);
```
dghelm, line 106: I'm wondering if we need to test for `encodedPK` potentially being passed in here ('%3A' instead of ':'). Maybe if someone is doing in-browser URL parsing?

src/registry.ts
```diff
@@ -164,6 +165,9 @@ export function getEntryUrl(
     ...customOptions,
   };
 
+  // Trim the prefix if it was passed in.
+  publicKey = trimPrefix(publicKey, "ed25519:");
```
dghelm, line 180: If worried about URL encoded public key passed in, using `decodeURI(publicKey)`.

dghelm (dismissed):
Looks good! I'm not sure if passing in a URL encoded key is a use-case, but we may account and test for it, just in case. I'm good either way though.

src/registry.ts
mrcnski: I would only accept un-encoded input, trying to decode as a URI here doesn't feel right. But I can add a check for valid hex-encoded input and throw if not -- this would catch any unexpected prefixes or encodings (since `:` and `%` are not hex characters) and the user would be notified of the problem.

src/utils.ts
```diff
@@ -339,9 +339,27 @@ export function stringToUint8Array(str: string): Uint8Array {
...
+}
+
+/**
+ * Returns true if the input is a valid hex-encoded string.
+ *
+ * @param str - The input string.
+ * @returns - True if the input is hex-encoded.
+ */
+export function isHexString(str: string): boolean {
+  return /^[0-9A-Fa-f]*$/g.test(str);
```
interstella-5555, line 362: Should it also pass on empty string? If not, you need to replace asterisk with plus.

```suggestion
  return /^[0-9A-Fa-f]+$/g.test(str);
```

interstella-5555 (approved):

src/utils.ts
mrcnski: I think it should pass on empty strings, so that the caller can handle that case differently (e.g. instead of throwing an "invalid hex string" error, can throw "empty/missing input")

dghelm (approved):

PR #226, opened 2020-12-08
Only return nulls if registry entry is nonexistent 

### LINKED ISSUE #220: getJSON and getEntry should throw on real errors
Right now, we consider any rejected response from api to be "empty record" for both getEntry and getJSON due to the fact that we had to reject it ourselves on timeout and couldn't rely on a specific error code from siad. Now that https://github.com/NebulousLabs/skynet-js/pull/218 implements timeout properly, we should be returning empty records only in an event when siad rejects with the not found error, otherwise we should be rethrowing the original api error.

mrcnski:
- Only return `null` from `getEntry` if registry entry is nonexistent

- Adds some checks on data returned from GET calls 

- Also adds typechecking to most of the exported functions. It is still missing some checks for objects but I got all the low-hanging fruit.

This should give more user-friendly errors than before, with actionable feedback. For example, calling `genKeyPairFromSeed` without the seed input would previously throw `Uncaught TypeError: Cannot read property 'length' of undefined`. Now it should throw `Expected parameter seed to be type string, was type undefined`.

Closes #220

### DIFF
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -90,17 +90,7 @@ export async function getEntry(
   dataKey: string,
   customOptions?: CustomGetEntryOptions
 ): Promise<SignedRegistryEntry> {
-  /* istanbul ignore next */
-  if (typeof publicKey !== "string") {
-    throw new Error(`Expected parameter publicKey to be type string, was type ${typeof publicKey}`);
-  }
-  /* istanbul ignore next */
-  if (typeof dataKey !== "string") {
-    throw new Error(`Expected parameter dataKey to be type string, was type ${typeof dataKey}`);
-  }
-  if (!isHexString(publicKey)) {
-    throw new Error("Expected parameter publicKey to be a hex-encoded string");
-  }
+  // Validation is done in `getEntryUrl`.
 
   const opts = {
     ...defaultGetEntryOptions,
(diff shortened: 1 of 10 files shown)


src/registry.ts
```diff
@@ -106,19 +119,30 @@ export async function getEntry(
...
+    // @ts-expect-error TS complains about err.response
+    throw new Error(err.response.data.message);
   }
 
-  if (response.status !== 200) {
-    return { entry: null, signature: null };
+  if (
+    typeof response.data.data != "string" ||
+    typeof response.data.revision != "string" ||
+    typeof response.data.signature != "string"
```
interstella-5555:
```suggestion
    typeof response.data.data !== "string" ||
    typeof response.data.revision !== "string" ||
    typeof response.data.signature !== "string"
```

src/registry.ts
```diff
@@ -106,19 +119,30 @@ export async function getEntry(
...
+  ) {
+    throw new Error("Did not get a complete entry response");
```
interstella-5555: Is this even possible to generate? Would this error tell user or developer how to proceed?

interstella-5555 (changes requested):

src/registry.ts
mrcnski: It's better than `type error: data.revision undefined` or something like that. I'll try to elaborate though

PS a user reported this so I guess it is possible.

interstella-5555 (dismissed):

interstella-5555 (approved):

ro-tex (approved):

PR #227, opened 2020-12-09
getFileContent should return a string and not object

mrcnski:
Previously, getFileContent would return a string for plaintext files and an
object for json files. It should only return a string -- in the case of json
files it will stringify the json.

### DIFF
--- a/src/download.ts
+++ b/src/download.ts
@@ -234,29 +234,25 @@ export async function getMetadata(
  * @returns - The content of the file.
  * @throws - Will throw if the skylinkUrl does not contain a skylink or if the path option is not a string.
  */
-export async function getFileContent(
+export async function getFileContent<T = unknown>(
   this: SkynetClient,
   skylink: string,
   customOptions?: CustomDownloadOptions
-): Promise<string> {
+): Promise<T> {
   const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
   const url = this.getSkylinkUrl(skylink, opts);
 
   // GET request the skylink
-  let { data } = await this.executeRequest({
+  const { data } = await this.executeRequest({
     ...opts,
     method: "get",
     url,
   });
 
-  if (typeof data !== "string") {
-    if (typeof data === "object" && data !== null) {
-      data = JSON.stringify(data);
-    } else {
-      throw new Error(
-        "Did not get 'data' in response despite a successful request. Please try again and report this issue to the devs if it persists."
-      );
-    }
+  if (typeof data === "undefined") {
+    throw new Error(
+      "Did not get 'data' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+    );
   }
 
   return data;
--- a/src/download.ts
+++ b/src/download.ts
@@ -238,18 +238,28 @@ export async function getFileContent(
   this: SkynetClient,
   skylink: string,
   customOptions?: CustomDownloadOptions
-): Promise<Record<string, unknown>> {
+): Promise<string> {
   const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
   const url = this.getSkylinkUrl(skylink, opts);
 
   // GET request the skylink
-  const response = await this.executeRequest({
+  let { data } = await this.executeRequest({
     ...opts,
     method: "get",
     url,
   });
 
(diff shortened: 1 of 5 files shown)


interstella-5555 (commented):
What is the motivation for this change? I assumed that this is correct and expected behaviour when @dhelma implemented this.

mrcnski:
I thought it was strange that `getFileContent` was returning different types. I changed it so that it returns the `unknown` type. Theres a few different possibilities, i.e. a zip file would be an arraybuffer, plaintext file a string, json file an object. There's not much we can do apart from return `unknown`.

dghelm:
I'm a bit unsure about the cleanest way of doing this.

Axios types data as `any,` but my limited TS knowledge makes me lean to using `unknown`. I like `unknown` since it's non-breaking.

But, disagreeing with my own code here -- it seems like returning an object with `content-type` and `data` fields (maybe even `Skynet-File-Metadata`) may allow more consistent expectations and clean code on the developer side. Applications wouldn't need to do checks using `typeof` to confirm data formats or anything, just rely on `content-type` since that determines how data was parsed and typed.

This also gives us room to add more fields in the future without breaking things -- this could especially be important for apps to get feedback from the server passively with requests once things like monetization are around.

dghelm:
Would it be too weird to setup an interface for this, with optional fields, so that HEAD and GET requests could maintain similar formats and we could formalize the format for axios data that ever makes it to the developer?

interstella-5555:
> Axios types data as `any,` but my limited TS knowledge makes me lean to using `unknown`. I like `unknown` since it's non-breaking.

Axios uses generics to type the response data and I think that's fair to assume user would know whether to type it with `any`, `string`, `object` or anything else.

mrcnski:
> But, disagreeing with my own code here -- it seems like returning an object with content-type and data fields (maybe even Skynet-File-Metadata) may allow more consistent expectations and clean code on the developer side. Applications wouldn't need to do checks using typeof to confirm data formats or anything, just rely on content-type since that determines how data was parsed and typed.

@dhelma I feel like this is overkill for this particular method since most of the time someone will just want to get file contents in a simple manner. Maybe you disagree based on talking to developers in which case let me know 🙂

However I think we should introduce a new function that returns content + headers -- e.g. `getFileContentRequest`, to mirror `uploadFileRequest`. It would return the full response object.

(Also, `uploadFileRequest` should be changed to return the full response object, like in the Python SDK.)

src/download.ts
```diff
@@ -234,22 +234,28 @@ export async function getMetadata(
...
   });
 
-  return response.data;
+  if (typeof data === "undefined") {
+    throw new Error(
+      "Did not get 'data' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+    );
+  }
+
+  return data;
```
ro-tex, line 258: Would it make sense to take this one step further and verify that `data` is `string`? Or we want to keep this intentionally open and let the caller make the assertion, so we can potentially add other data types as well?

src/skydb.ts
```diff
@@ -49,9 +49,12 @@ export async function getJSON(
   }
 
   // Download the data in that Skylink.
-  // TODO: Replace with download request method.
   const skylink = entry.data;
-  const data = await this.getFileContent(skylink, opts);
+  const data = await this.getFileContent<Record<string, unknown>>(skylink, opts);
```
ro-tex, line 53: Doesn't `getFileContent` return `Record<unknown>` now and not `Record<string, unknown>`?

ro-tex (approved):
LGTM.

Two comments that are both "I might be wrong but please take a look at this". Feel free to resolve them if the answer is "nah, it's OK".

src/download.ts
ro-tex: Maybe I should have read the existing discussion before reading the code. https://github.com/NebulousLabs/skynet-js/pull/227#issuecomment-742692870 resolves this.

src/skydb.ts
mrcnski: Thanks for the review. We had a discussion and decided to have it return generic data, since depending on the file it can return a string, json, or arraybuffer (e.g. binary files). We're going to also return the content-type header that can be examined.

mrcnski:
Closing in flavor of #239

PR #228, opened 2020-12-11
[BREAKING] make upload request functions return full response

mrcnski:
The upload request functions were previously not that useful. The intended purpose of these functions was to get the full response so that e.g. headers can be examined.


mrcnski:
After internal discussion we decided to remove the request functions instead.

PR #239, opened 2020-12-15
[BREAKING] Refactor getfilecontent and upload methods

mrcnski:
- [BREAKING] Upload methods return object
- [BREAKING] Remove upload request
- [BREAKING] Make getMetadata return object containing metadata and skylink
- Return object containing contentType, metadata, skylink from getFileContent
- getFileContent returns a skylink prefixed with sia:
- Add getFileContentHns

### DIFF
--- a/src/download.ts
+++ b/src/download.ts
@@ -52,6 +52,17 @@ export type GetFileContentResponse<T = unknown> = {
 };
 
 /**
+ * The response for a get metadata request.
+ *
+ * @property metadata - The metadata in JSON format.
+ * @property skylink - 46-character skylink.
+ */
+export type GetMetadataResponse = {
+  metadata: Record<string, unknown>;
+  skylink: string;
+};
+
+/**
  * The response for a resolve HNS request.
  *
  * @property skylink - 46-character skylink.
@@ -226,7 +237,7 @@ export async function getMetadata(
   this: SkynetClient,
   skylinkUrl: string,
   customOptions?: CustomDownloadOptions
-): Promise<Record<string, unknown>> {
+): Promise<GetMetadataResponse> {
   const opts = { ...defaultDownloadOptions, ...this.customOptions, ...customOptions };
   const url = this.getSkylinkUrl(skylinkUrl, opts);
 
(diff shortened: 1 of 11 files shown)


src/download.ts
```diff
@@ -36,6 +37,32 @@ export type CustomHnsDownloadOptions = CustomDownloadOptions & {
...
+/**
+ * The response for a get metadata request.
+ *
+ * @property metadata - The metadata in JSON format.
+ * @property skylink - 46-character skylink.
+ */
+export type GetMetadataResponse = {
+  metadata: Record<string, unknown>;
+  skylink: string;
+};
```
dghelm: Should we combine these return types with `data?: T` ? contentType is returned on HEAD calls and could serve a purpose.

mrcnski: I want to keep the response types separate, but I added contentType to the metadata response. I'm can see potential use cases for it

dghelm (approved):

src/download.ts
```diff
@@ -277,14 +316,70 @@ export async function getFileContent(
...
+      "Did not get 'data' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+    );
+  }
+  if (typeof response.headers === "undefined") {
+    throw new Error(
+      "Did not get 'headers' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+    );
+  }
+
+  // TODO: Return null instead if header not found?
```
ro-tex, line 377: IMO an empty content type is acceptable, especially if we're doing it for the metadata call. It's better than `null` because the client can treat it as a string in a safe way without needing a nullish check.

src/integration.test.ts
```diff
@@ -105,14 +105,26 @@ describe("Registry end to end integration tests", () => {
...
+    const { data } = await client.getFileContent(skylink);
+    expect(data).toEqual(expect.any(String));
+    expect(data).toEqual(data);
+  });
+
+  it("Should get JSON file contents", async () => {
     const json = { key: "testdownload" };
 
     // Upload the data to acquire its skylink
     const file = new File([JSON.stringify(json)], dataKey, { type: "application/json" });
```
ro-tex, line 124: What happens if we don't specify the content type? Would it be detected?
No need to cover this now but it might be interesting to verify the behaviour.

ro-tex (approved):
LGTM!

src/integration.test.ts
mrcnski: Follow-up: https://github.com/NebulousLabs/skynet-js/issues/241

PR #242, opened 2020-12-18
Version update to 3.0.0-beta

mrcnski:


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## [3.0.0-beta]
```
dghelm, line 3: Should we do `beta1` just in case?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "skynet-js",
-  "version": "2.9.0",
+  "version": "3.0.0-beta",
```
dghelm, line 3: `beta1`?

dghelm (dismissed):
Should we number beta releases in case we want to do a few modifications? If not, this looks great.

CHANGELOG.md
mrcnski: Would that make people expect more betas? If they see `beta1` they may think `beta2` is coming. Since we only plan on having one (unless I missed something) I would stay with `beta` for now

dghelm: That's true. I think it'd just be in case there are problems with the release that small changes could be numbered. `-beta.0` might be cleared for showing that, but we could also just number afterwards.

dghelm: I'm wrong -- sorry. Semver says

> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
https://semver.org/

PR #266, opened 2021-01-13
Add integration tests for full file contents and metadata

### LINKED ISSUE #241: Follow-up about content type
From https://github.com/NebulousLabs/skynet-js/pull/239#discussion_r545863303:

> What happens if we don't specify the content type? Would it be detected?
No need to cover this now but it might be interesting to verify the behaviour.

mrcnski:
**Note:** Tests currently failing due to access controls on siasky.net blocking the skynet-skylink header.

Closes #241

### DIFF
--- a/src/integration.test.ts
+++ b/src/integration.test.ts
@@ -104,17 +104,49 @@ describe("Registry end to end integration tests", () => {
   });
 });
 
-describe("Upload and download integration tests", () => {
+describe("Upload and download end-to-end tests", () => {
+  const fileData = "testing";
+  const plaintextType = "text/plain";
+  const plaintextMetadata = {
+    filename: dataKey,
+    length: fileData.length,
+    subfiles: {
+      HelloWorld: { filename: dataKey, contenttype: plaintextType, len: fileData.length },
+    },
+  };
+
   it("Should get plaintext file contents", async () => {
-    const fileData = "testing";
+    // Upload the data to acquire its skylink.
 
-    // Upload the data to acquire its skylink
-    const file = new File([fileData], dataKey, { type: "text/plain" });
+    const file = new File([fileData], dataKey, { type: plaintextType });
     const { skylink } = await client.uploadFile(file);
+    expect(skylink).not.toEqual("");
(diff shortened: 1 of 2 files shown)


ro-tex (dismissed):

src/integration.test.ts
```diff
@@ -3,7 +3,7 @@ import { MAX_GET_ENTRY_TIMEOUT } from "./registry";
 import { MAX_REVISION } from "./utils";
 
 // TODO: Use siasky.dev when available.
```
ro-tex: This line can go now. ;)

src/integration.test.ts
```diff
@@ -150,15 +151,30 @@ describe("Upload and download end-to-end tests", () => {
...
+    const { skylink } = await client.uploadFile(file);
+    expect(skylink).not.toEqual("");
+
+    // Get file content and check returned values.
+
+    const { data, contentType } = await client.getFileContent(skylink);
 
     expect(data).toEqual(expect.any(Object));
     expect(data).toEqual(json);
+    expect(contentType).toEqual("application/octet-stream");
```
ro-tex: What happens if we specify the file's name and it has an extension? The content type should be guessed correctly, right? Can we add a test for that as well?

Actually, if the content allows it, we should get some content type heuristics even without a filename. All that "will read up to X bytes and guess the content" magic. Doesn't that work or is this JSON too short?

ro-tex (dismissed):
LGTM. I left some comments but nothing is blocking.

src/integration.test.ts
mrcnski: Looks like siasky.dev is still not working (the registry is broken), going back to siasky.net for now

src/integration.test.ts
mrcnski: I don't know why it doesn't infer from the content, but I added a test case for inferring from filename

ro-tex (dismissed):

dghelm (approved):

ro-tex (approved):

PR #269, opened 2021-01-14
Fix eslint warnings

mrcnski:
Also any warnings will cause CI to fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "test": "jest --coverage --coverageDirectory ../coverage",
     "build": "rimraf dist && babel src --out-dir dist --extensions .ts --ignore src/**/*.test.ts && tsc --project tsconfig.build.json",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "test": "jest --coverage --coverageDirectory ../coverage",
     "build": "rimraf dist && babel src --out-dir dist --extensions .ts --ignore src/**/*.test.ts && tsc --project tsconfig.build.json",
-    "lint:eslint": "eslint --ext .ts utils src",
+    "lint:eslint": "eslint --ext .ts utils src --max-warnings 0",
```
ro-tex, line 17: Hardcore. I like it! :)

ro-tex (approved):

package.json
mrcnski: Haha. This basically just treats warnings as errors. Ensures we don't miss anything

peterjan (approved):

PR #270, opened 2021-01-15
Add support for no-response-metadata

### LINKED ISSUE #268: Add download option to support no-response-metadata
The download options object should be expanded with a new option to pass the `no-response-metadata` HTTP parameter

mrcnski:
Closes #268

### DIFF
--- a/src/download.test.ts
+++ b/src/download.test.ts
@@ -131,6 +131,12 @@ describe("getSkylinkUrl", () => {
     expect(url).toEqual(`${expectedUrl}?no-response-metadata=true`);
   });
 
+  it("should return correctly formed URLs with no-response-metadata set and with forced download", () => {
+    const url = client.getSkylinkUrl(skylink, { download: true, noResponseMetadata: true });
+
+    expect(url).toEqual(`${expectedUrl}?attachment=true&no-response-metadata=true`);
+  });
+
   const expectedBase32 = `https://${skylinkBase32}.siasky.net`;
 
   it.each(validSkylinkVariations)("should convert base64 skylink to base32 using skylink %s", (fullSkylink) => {
--- a/src/download.test.ts
+++ b/src/download.test.ts
@@ -85,6 +85,12 @@ describe("getHnsUrl", () => {
 
     expect(url).toEqual(`${expectedHnsUrl}${attachment}`);
   });
+
+  it("should return correctly formed hns URL with no-response-metadata set", () => {
+    const url = client.getHnsUrl(hnsLink, { noResponseMetadata: true });
+
+    expect(url).toEqual(`${expectedHnsUrl}?no-response-metadata=true`);
+  });
 });
 
 describe("getHnsresUrl", () => {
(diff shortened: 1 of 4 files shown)


src/download.test.ts
```diff
@@ -114,9 +120,15 @@ describe("getSkylinkUrl", () => {
...
-    );
+    const url = client.getSkylinkUrl(skylink, { download: true, path: "foo?bar" });
+
+    expect(url).toEqual(`${expectedUrl}/foo%3Fbar${attachment}`);
+  });
+
+  it("should return correctly formed URLs with no-response-metadata set", () => {
+    const url = client.getSkylinkUrl(skylink, { noResponseMetadata: true });
+
+    expect(url).toEqual(`${expectedUrl}?no-response-metadata=true`);
```
ro-tex, line 131: very nitty nit: For composition cases like this one `toContain` might be better than `toEqual` but I don't think we need to change it right now. It might be helpful in case where we add default query params or something similar.

ro-tex (dismissed):

src/download.test.ts
```diff
@@ -114,9 +120,15 @@ describe("getSkylinkUrl", () => {
...
   it("should return correctly formed URLs with forced download and path", () => {
-    expect(client.getSkylinkUrl(skylink, { download: true, path: "foo?bar" })).toEqual(
-      `${expectedUrl}/foo%3Fbar${attachment}`
-    );
+    const url = client.getSkylinkUrl(skylink, { download: true, path: "foo?bar" });
+
+    expect(url).toEqual(`${expectedUrl}/foo%3Fbar${attachment}`);
+  });
+
+  it("should return correctly formed URLs with no-response-metadata set", () => {
```
peterjan, line 128: properly also want to test a case where there's multiple params added to the query string?

src/download.ts
```diff
@@ -159,13 +161,17 @@ export function getSkylinkUrl(this: SkynetClient, skylinkUrl: string, customOpti
   if (opts.download) {
     query.attachment = true;
   }
+  if (opts.noResponseMetadata) {
+    query["no-response-metadata"] = true;
```
peterjan, line 167: why are we deviating from the format here? `query.attachment` vs `query["no-response-metadata"]
shouldn't we use camel casing? `query.noResponseMetadata`?

peterjan (dismissed):

interstella-5555 (dismissed):

src/download.test.ts
mrcnski: I would argue that `toEqual` is generally better as it catches any changes to the value. Then we have the chance to confirm that the change was desired and update the test if so. You make a good point about default query params, we can address that when/if we get to it.

src/download.ts
mrcnski: Well the parameter is `no-response-metadata`, camel casing wouldn't work.

src/download.test.ts
mrcnski: Also I think `toContain` wouldn't quite work unless no-response-metadata (in this example) was always first. We'd have to come up with something else, maybe even a custom matcher.

ro-tex: `.toContain("no-response-metadata=true")` :D

I agree with the rest of your point. Definitely no need to change anything now.

src/download.test.ts
```diff
@@ -131,6 +131,12 @@ describe("getSkylinkUrl", () => {
     expect(url).toEqual(`${expectedUrl}?no-response-metadata=true`);
   });
 
+  it("should return correctly formed URLs with no-response-metadata set and with forced download", () => {
+    const url = client.getSkylinkUrl(skylink, { download: true, noResponseMetadata: true });
+
+    expect(url).toEqual(`${expectedUrl}?attachment=true&no-response-metadata=true`);
```
ro-tex, line 137: I think these might appear in a different order and randomly fail the test. IIRC, ordering is not guaranteed when serializing objects in JS and this is what essentially happens in `addUrlQuery`:
```js
  query = { ...parsed.query, ...query };
  parsed.set("query", query);
  return parsed.toString();
```
I might be wrong, though.

In any case, not a blocker.

ro-tex (approved):

interstella-5555 (approved):

PR #277, opened 2021-01-22
Run integration tests daily

mrcnski:
1. The daily run will test against all `*.siasky.net` servers
2. ~~The PR CI pipeline will use `siasky.dev` instead of `siasky.net`~~ Use `siasky.net` since dev has a broken registry.
3. Individual servers can be tested by e.g. `SKYNET_JS_INTEGRATION_TEST_SERVER=https://eu-fin-1.siasky.net yarn run jest src/integration.test.ts`

Example successful run:
https://github.com/NebulousLabs/skynet-js/pull/277/checks?check_run_id=1768989935

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -19,7 +19,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Daily CI
+      - name: Use Node.js
         uses: actions/setup-node@v1
-      - run: yarn
-      - run: SKYNET_JS_INTEGRATION_TEST_SERVER="https://${{ matrix.server }}.siasky.net" yarn run jest src/integration.test.ts
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Integration test on server
+        run: yarn run jest src/integration.test.ts
+        env:
+          SKYNET_JS_INTEGRATION_TEST_SERVER: "https://${{ matrix.server }}.siasky.net"
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -22,4 +22,4 @@ jobs:
       - name: Daily CI
         uses: actions/setup-node@v1
       - run: yarn
-      - run: SKYNET_JS_INTEGRATION_TEST_SERVER="https://${{ matrix.server }}.siasky.net" yarn test src/integration.test.ts
+      - run: SKYNET_JS_INTEGRATION_TEST_SERVER="https://${{ matrix.server }}.siasky.net" yarn run jest src/integration.test.ts
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,0 +1,17 @@
+# This workflow will run integrations tests daily at midnight.
+
+name: Daily CI
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Daily CI
+        uses: actions/setup-node@v1
+      - run: ./test-all-servers
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -3,9 +3,6 @@
 name: Daily CI
 
 on:
-  # JUST FOR TESTING THE WORKFLOW
-  pull_request:
-    branches: [master]
   schedule:
     - cron: "0 0 * * *"
 
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -3,15 +3,23 @@
 name: Daily CI
 
 on:
+  # JUST FOR TESTING THE WORKFLOW
+  pull_request:
+    branches: [master]
   schedule:
     - cron: "0 0 * * *"
 
 jobs:
   test:
+    name: Daily Server Test
     runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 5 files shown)


ro-tex (dismissed):

test-all-servers
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# declare an array with all servers
+array=( eu-ger-1 eu-ger-2 eu-fin-1 us-or-1 us-or-2 us-pa-1 us-pa-2 us-va-1 )
+
+for SERVER in "${array[@]}"
+do
+	SKYNET_JS_INTEGRATION_TEST_SERVER="https://$SERVER.siasky.net" yarn test src/integration.test.ts
+done
```
interstella-5555: You should use matrix strategy for listing test servers instead of this manual loop.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

interstella-5555 (changes requested):

test-all-servers
mrcnski: From what i understand we want to run this script from the healthcheck as well.

interstella-5555: I don't think there is any reason to do so, health checks run on each server and test only that specific server's health, we don't want to run a full test suite against all servers from every single server

mrcnski: Okay, I'll remove this script and make a follow-up issue for the healthcheck

mrcnski: See https://github.com/NebulousLabs/skynet-webportal/issues/598

.github/workflows/daily.yml
```diff
@@ -0,0 +1,25 @@
...
+    strategy:
+      matrix:
+        server: [eu-ger-1, eu-ger-2, eu-fin-1, us-or-1, us-or-2, us-pa-1, us-pa-2, us-va-1]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Daily CI
+        uses: actions/setup-node@v1
+      - run: yarn
+      - run: SKYNET_JS_INTEGRATION_TEST_SERVER="https://${{ matrix.server }}.siasky.net" yarn run jest src/integration.test.ts
```
interstella-5555: It looks cleaner when you put dependencies installation in its separate step and use `env` for environment variables.

```suggestion
      - uses: actions/checkout@v2
      - name: Use Node.js
        uses: actions/setup-node@v1

      - name: Install dependencies
        run: yarn
        
      - name: Daily CI
        run: yarn run jest src/integration.test.ts
        env:
          SKYNET_JS_INTEGRATION_TEST_SERVER: "https://${{ matrix.server }}.siasky.net"
```

src/integration.test.ts
```diff
@@ -2,12 +2,18 @@ import { genKeyPairAndSeed, SkynetClient } from "./index";
...
 
-// TODO: Use siasky.dev when available.
-const client = new SkynetClient("https://siasky.net");
+let portal = "https://siasky.net";
+
+// To test a specific server, e.g. SKYNET_JS_INTEGRATION_TEST_SERVER=https://eu-fin-1.siasky.net yarn test src/integration.test.ts
+const testServer = process.env.SKYNET_JS_INTEGRATION_TEST_SERVER;
+if (testServer) {
+  portal = testServer;
+}
```
interstella-5555:
```suggestion
// To test a specific server, e.g. SKYNET_JS_INTEGRATION_TEST_SERVER=https://eu-fin-1.siasky.net yarn test src/integration.test.ts
const portal = process.env.SKYNET_JS_INTEGRATION_TEST_SERVER || "https://siasky.net";
```

src/integration.test.ts
```diff
@@ -2,12 +2,18 @@ import { genKeyPairAndSeed, SkynetClient } from "./index";
...
+console.log(`Running integration tests with portal ${portal}`);
```
interstella-5555: instead of console.log, I would wrap all of the tests in a describe block and put it there like

```javascript
describe("Integration test for ${portal}", () => {
  // all other "describe" copied here
});
```

interstella-5555 (commented):
mostly cosmetic comments, nothing really blocking 👍

interstella-5555 (dismissed):

.github/workflows/daily.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: Daily Server Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        server: [eu-ger-1, eu-ger-2, eu-fin-1, us-or-1, us-or-2, us-pa-1, us-pa-2, us-va-1]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
```
ro-tex, line 20: It would be nice to specify the version of Node we want thin run on.

e.g.:
```suggestion
      - uses: actions/checkout@v2
      - name: Use Node.js
      - uses: actions/setup-node@v1
        with:
            node-version: '14'
```

source: https://github.com/actions/setup-node

ro-tex (dismissed):
LGTM.

Left a comment but it's optional.

.github/workflows/daily.yml
mrcnski: For some reason I thought it would run the latest version. However, looking at the successful run in the OP it looks like it picked v10. Will change it.

ro-tex (approved):

interstella-5555 (approved):

PR #305, opened 2021-02-12
use Promise.all to run first setJSON requests in parallel

dghelm:
Initial draft for running first two requests in parallel

### DIFF
--- a/src/skydb.ts
+++ b/src/skydb.ts
@@ -139,11 +139,12 @@ export async function setJSON(
       throw new Error("Current entry already has maximum allowed revision, could not update the entry");
     }
   } else {
-    // Assert the input is 64 bits.
-    const [skyfile] = await Promise.all<UploadRequestResponse>([skyfilePromise]);
+    const skyfile = await skyfilePromise;
     skylink = skyfile.skylink;
-    assertUint64(revision);
   }
+
+  // Assert the input is 64 bits.
+  assertUint64(revision);
 
   // build the registry value
   const entry: RegistryEntry = {
--- a/src/skydb.ts
+++ b/src/skydb.ts
@@ -11,7 +11,7 @@ import {
   isHexString,
 } from "./utils";
 import { Buffer } from "buffer";
-import { CustomUploadOptions } from "./upload";
+import { CustomUploadOptions, UploadRequestResponse } from "./upload";
 import { CustomDownloadOptions } from "./download";
 
 /**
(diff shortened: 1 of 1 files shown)


dghelm:
Have checked code in-browser -- works well:
image

src/skydb.ts
```diff
@@ -124,13 +140,11 @@ export async function setJSON(
     }
   } else {
     // Assert the input is 64 bits.
+    const [skyfile] = await Promise.all<UploadRequestResponse>([skyfilePromise]);
```
mrcnski: Don't need `Promise.all` here, can just `await`.

src/skydb.ts
```diff
@@ -106,12 +106,28 @@ export async function setJSON(
...
-    const entry: SignedRegistryEntry = await this.registry.getEntry(toHexString(publicKey), dataKey, opts);
+    // start getEntry, do not block.
+    const entryPromise: Promise<SignedRegistryEntry> = this.registry.getEntry(toHexString(publicKey), dataKey, opts);
+
+    // Block until both getEntry and Skyfile upload are finished.
+    const [entry, skyfile] = await Promise.all<SignedRegistryEntry, UploadRequestResponse>([
+      entryPromise,
+      skyfilePromise,
+    ]);
+    skylink = skyfile.skylink;
```
mrcnski, line 129: I think we can move this line out of the `if` and `else` branches, it's not needed until later and would save 1 LoC.

src/skydb.ts
```diff
@@ -124,13 +140,11 @@ export async function setJSON(
     }
   } else {
     // Assert the input is 64 bits.
```
mrcnski: Looks like the comment is in the wrong spot now. However, I'd move the assertion out of the `else` clause, after the `if-else` blocks. We might as well assert the `revision` calculated in the `if`, even though it should never fail.

mrcnski (dismissed):
Looks good! Some minor comments

src/skydb.ts
ro-tex: I was wondering why we're declaring the var at line 114 but I guess we can avoid that and just declare it with its init around L147?

ro-tex (dismissed):
LGTM, nothing blocking.

src/skydb.ts
dghelm: This may be me not understanding TS, but I get errors if I try to reference skyfile outside the if/else block since it was initialized inside it. This is the motivation for the weird `let skylink: string;` above.

ro-tex: ~~Oh, yes, your approach is absolutely correct - if you want to use the var in the block and then outside of it you need to declare it outside. But, as Marcin pointed out, you can just do it after the `if` block. That is, if I am reading the code correctly.~~

If you declare `skyfile` outside and then after the `if` blocks you do
```js
let skylink = skyfile.skylink;
```
that should do the trick, IMO.

It's not a big deal in any case.

dghelm: https://www.typescriptlang.org/docs/handbook/variable-declarations.html#block-scoping

ro-tex (approved):

mrcnski (approved):

src/skydb.ts
mrcnski: I think what @ro-tex suggested with declaring `skyfile` before the scope should work. But it's such a minor thing that I'm just going to merge this :)

dghelm: Didn't see your post before putting the link there! But still, I get "Cannot find name: skyfile" because it was declared in the blockscope of the if statement.

dghelm: I'm too slow! I was missing an array restructuring assignment thing, but I'll do another pull. Thanks for the insight!

PR #319, opened 2021-02-15
Remove node-forge library

mrcnski:
This PR gets the final bundled size of skynet-js down from 2.7mb to 1.1 mb by removing `node-forge` for crypto and replacing it with `tweetnacl-js` and `sjcl`.

`node-forge` did not use ES6 export style, and so the entire library was bundled in even though we used only a few of its functions:

https://github.com/digitalbazaar/forge/issues/609

## Compatibility

We already had backwards compatibility tests for affected crypto functions, so this change will not break anything.

### Before

Screen Shot 2021-02-13 at 10 53 14

### After

Screen Shot 2021-02-14 at 21 16 08


ro-tex (dismissed):

interstella-5555 (dismissed):

interstella-5555 (approved):

dghelm (approved):
Seems good to me! Great find, and cool to be moving to `strict: true` tsconfig.

PR #322, opened 2021-02-17
Downgrade husky to major version 4

mrcnski:
As discussed on Discord `husky` uses a Open Parity license that is incompatible with our project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "sjcl": "^1.0.8",
     "tweetnacl": "^1.0.3",
     "url-join": "^4.0.1",
-    "url-parse": "=1.4.7"
+    "url-parse": "1.4.7"
   },
   "devDependencies": {
     "@babel/cli": "^7.11.6",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
       "eslint --ext .ts --fix",
-      "tsc --esModuleInterop --noemit",
+      "tsc --esModuleInterop",
       "prettier --write"
     ],
     "*.{json,yml,md}": [
@@ -100,7 +100,7 @@
     "eslint": "^7.11.0",
     "eslint-plugin-compat": "^3.8.0",
     "eslint-plugin-jsdoc": "^32.0.0",
-    "husky": "^5.0.9",
+    "husky": "^4.3.8",
     "jest": "^26.4.2",
     "lint-staged": "^10.3.0",
     "prettier": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "eslint": "^7.11.0",
     "eslint-plugin-compat": "^3.8.0",
     "eslint-plugin-jsdoc": "^32.0.0",
-    "husky": "=4.3.8",
+    "husky": "4.3.8",
     "jest": "^26.4.2",
     "lint-staged": "^10.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
       "eslint --ext .ts --fix",
-      "tsc --esModuleInterop --noemit",
+      "tsc --esModuleInterop",
```
interstella-5555, line 45: should this be included in the downgrade? is this interconnected?

mrcnski: No harm in having it here, it's a minor change. tsconfig.json already has noemit. I can also just remove this change, doesn't matter either way.

interstella-5555 (dismissed):

mrcnski:
I have just downgraded `url-parse` as well. Update `1.4.7 -> 1.5.0` was released 21 hours ago and it broke our tests. I don't have time to go fix all these tests just because some dependency doesn't respect semver so I set it to a fixed version.

package.json
```diff
@@ -92,15 +92,15 @@
...
+    "@types/url-parse": "=1.4.3",
     "@typescript-eslint/eslint-plugin": "^4.3.0",
     "@typescript-eslint/parser": "^4.3.0",
     "axios-mock-adapter": "^1.18.2",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "eslint": "^7.11.0",
     "eslint-plugin-compat": "^3.8.0",
     "eslint-plugin-jsdoc": "^32.0.0",
-    "husky": "^4.3.8",
+    "husky": "=4.3.8",
```
interstella-5555:
```suggestion
    "husky": "4.3.8",
```

package.json
```diff
@@ -92,15 +92,15 @@
     "@types/randombytes": "^2.0.0",
     "@types/sjcl": "^1.0.29",
     "@types/url-join": "^4.0.0",
-    "@types/url-parse": "^1.4.3",
+    "@types/url-parse": "=1.4.3",
```
interstella-5555:
```suggestion
    "@types/url-parse": "1.4.3",
```

package.json
```diff
@@ -76,7 +76,7 @@
     "sjcl": "^1.0.8",
     "tweetnacl": "^1.0.3",
     "url-join": "^4.0.1",
-    "url-parse": "^1.4.7"
+    "url-parse": "=1.4.7"
```
interstella-5555:
```suggestion
    "url-parse": "1.4.7"
```

interstella-5555 (changes requested):
You don't need to use `=` in package.json to bind library to specific version, it's not really a common practice to do so too, haven't seen anyone do that actually. Otherwise it's 👍

interstella-5555:
You will need to rerun `yarn` to update the lock file too @m-cat

ro-tex:
Can we add a note at a visible place that explains that we need to stay at version 4 because of licensing? Otherwise we might accidentally upgrade in the future. Same for `url-parse`.

Maybe a dedicated `DEPENDENCIES.md`?

ro-tex (approved):

dghelm:
Will `skynet-js` need to use the same license as `skyd` or can we just keep it at MIT?

Edit: Confirmed with David that `skynet-js` can continue to be MIT, it's just `skyd` that will be Skynet Licensed

mrcnski:
Closing since this repo is staying on MIT, which should be fine with Parity License. Can reopen if needed.

PR #323, opened 2021-02-18
Update version to 3.0.1-beta

mrcnski:


mrcnski:
Merged without peer review as I already accidentally published it lol. Luckily it is a beta version and not stable 😉

PR #336, opened 2021-02-22
Change mime dep to reduce bundle size

### LINKED ISSUE #320: Replace mime-db dependency
Inspired by #319 I looked into `mime-db` -- it seems like the original usage was prompted by [an issue](https://github.com/NebulousLabs/skynet-webportal/issues/290#issuecomment-660184429) with `react-dropzone` on the web portal homepage.

This dependency is just a large JSON file of all known MIME-types. It seems like this would be a developer's responsibility to manage, and at 200kb unpacked, it's almost 20% of the space in our code.

**Option 1:** Fix `react-dropzone` and drop dependency.

**Option 2:** Follow their documentation's advice:

> If you're crazy enough to use this in the browser, you can just grab the JSON file using jsDelivr. It is recommended to replace master with a release tag as the JSON format may change in the future.

We could host the file on skynet and have the client load it from the current portal if an upload of unknown MIME-type ever occurs.

**Option 3:** Consider a library like [mime](https://www.npmjs.com/package/mime). It's a self-contained module bundled with a pre-optimized version of the mime-db dataset, and intended for browser use. `mime-db`, when minified and compressed, is 18KB, `mime` is 8KB, `mime/lite` is 2KB.

mrcnski:
Closes #320


dghelm (approved):
LGTM

ro-tex (approved):

PR #337, opened 2021-02-23
Check error response fields in getEntry

mrcnski:


src/registry.ts
```diff
@@ -1,4 +1,4 @@
-import { AxiosResponse } from "axios";
+import type { AxiosResponse } from "axios";
```
ro-tex, line 1: TIL...

src/registry.ts
```diff
@@ -119,12 +119,29 @@ export async function getEntry(
         return JSON.parse(data);
       },
     });
-  } catch (err: unknown) {
-    // @ts-expect-error TS complains about err.response
+  } catch (err) {
+    // Check if status was 404 "not found" and return null if so.
```
ro-tex: Comment should two `if`s down.

src/registry.ts
```diff
@@ -119,12 +119,29 @@ export async function getEntry(
...
+    /* istanbul ignore next */
+    if (!err.response.status) {
+      throw new Error("Error response did not contain expected field 'status'");
+    }
     if (err.response.status === 404) {
       return { entry: null, signature: null };
     }
-    // @ts-expect-error TS complains about err.response
+
+    // Else return the error message from the response.
```
ro-tex: This continues the logic of the previous comment. Chances are that the reader will be spot-reading the source of the error they're tracing, so they won't read the first comment before this one and they sit here scratching their head for a while. It's not a problem (they'll eventually get it) but it can be improved by rewording or removing the comment.

Also, should we log/return the entire error response in the case where it's missing the fields we expect? It could eb informative to do so but, otoh, it might also be useless to the dev - I'm not entirely sure.

ro-tex (commented):
LGTM, minor comments on the comments. :P

src/registry.ts
```diff
@@ -119,12 +119,29 @@ export async function getEntry(
...
+    /* istanbul ignore next */
+    if (!err.response.data) {
+      throw new Error("Error response did not contain expected field 'data'");
+    }
+    /* istanbul ignore next */
+    if (!err.response.data.message) {
+      throw new Error("Error response did not contained expected fields 'data.message'");
+    }
```
interstella-5555: I would include status code in those messages since it might be helpful and at this point you already checked that it is available.

interstella-5555 (changes requested):

src/registry.ts
mrcnski: Yes

interstella-5555 (approved):

ro-tex (approved):

PR #338, opened 2021-02-23
Build web, node, and production bundle

mrcnski:
I've moved to Webpack and created two separate targets for `node` and `dev`.

### Production build deploy

1. [x] The production web build should be deployed to our `skynet-js` HNS. I plan to include a `deploy.js` script in the repo which uses the Node version of the library itself to upload the production file and save the skylink to the registry. (#351)

### Update docs

1. [ ] Remove nodejs-skynet SDK from the docs (and archive repo)
2. [ ] Update nodejs examples
3. [ ] Add notes about differences in API between Node and Web (since the former lacks `File` and the latter lacks access to the local filesystem)


mrcnski:
Closing in flavor of https://github.com/SkynetHQ/skynet-js/pull/16

PR #352, opened 2021-03-02
Add parse.JSON in getJSON

mrcnski:


src/skydb.ts
```diff
@@ -59,8 +59,17 @@ export async function getJSON(
 
   // Download the data in that Skylink.
   const skylink = entry.data;
-  const { data } = await this.getFileContent<Record<string, unknown>>(skylink, opts);
-
+  let { data } = await this.getFileContent<Record<string, unknown>>(skylink, opts);
```
interstella-5555, line 62: `getFileContent` will return `object` on valid json, if it is string we should assume it's invalid, you shouldn't need to check it again here because axios already does that in the method that's used internally

interstella-5555 (changes requested):

src/skydb.ts
mrcnski: For background, this is the skylink that was causing a failure. It was created using an ~~old~~ *current* version of SkyID that stringified the JSON before calling `setJSON`:

https://siasky.net/AAA2Tj5k1Zw839VMUyfv_VmT3TRXUtc3lPQrVKWOuXTKlg

Doesn't hurt to support this case for backwards compatibility.

interstella-5555: That's not valid JSON though - it's a string. I don't think it's good idea to allow invalid input and call it valid, json format is well defined and axios does a good job parsing it there - we shouldn't try to correct it and thus keep allowing invalid formatting.

mrcnski: What is the harm of allowing this? This allows apps to seamlessly ~~upgrade from old~~ *interoperate with* SkyID without breakage.

> axios does a good job parsing it there

Well, no, hence why I parse it manually.

interstella-5555: The harm is that the behaviour is not correct since the output is not a an object - it is a string, the valid usage is that you would expect the string to be returned and not parsed as an object - https://www.ecma-international.org/publications-and-standards/standards/ecma-404/

> Well, no, hence why I parse it manually.

But the whole idea is that you shouldn't do that because axios correctly recognises the data format and this is not object but a string.

mrcnski: We need it to be compatible with SkyID unfortunately:

https://github.com/DaWe35/SkyID/blob/main/src/connect.js#L95

I still don't see how it's not correct. If it's not valid JSON, then `JSON.parse` fails.

interstella-5555: A string is technically a valid json, just not a json object - it's still a string according to the specification and it's not a correct usage to try and parse that string again because you can end up with all sorts of issues with different types. Also some people could deliberately stringify their values and expect a string when they pull it out. SkyID should fix the way it uploads the json object when they anyone using it upgrades to new version of skynet-js, not the other way around.

```javascript
> JSON.parse("0")
0
> typeof JSON.parse("0")
'number'
> JSON.parse("false")
false
> typeof JSON.parse("false")
'boolean'
> JSON.parse("[]")
[]
> typeof JSON.parse("[]")
'object'
```

mrcnski: > it's not a correct usage to try and parse that string again because you can end up with all sorts of issues with different types.

I don't understand the relevance of this, nor of the code exampels you provided. If the string fails to parse, we throw an error, If the parsed JSON is not an object, we throw an error.

> Also some people could deliberately stringify their values and expect a string when they pull it out.

I don't understand this either. The SDK does type-checking in `setJSON` now so it isn't even possible. In the case where someone was doing that with the old SDK, they are not able to update to the new SDK because it errors on retrieving existing data instead of just parsing the string as a JSON object internally when it can.

I don't see any good reason not to preserve data compatibility with previous versions of the SDK and with SkyID. At this point it is just nitpicking and I hope we can stop dragging on this pointless argument.

interstella-5555: You need to look at the raw http responses to understand the difference:

one:
```json
{"foo":"bar"}
```

two:
```json
"{\"foo\":\"bar\"}"
```

The difference between those is that first one is an **object** and second one is a **string**. First one should be represented in javascript as `object` and second one as a `string`.

If my opinion is not enough, think of the industry standards and leaders and how they handle [your example](https://siasky.net/AAA2Tj5k1Zw839VMUyfv_VmT3TRXUtc3lPQrVKWOuXTKlg):

- fetch:
Screenshot 2021-03-08 at 12 15 18
 
- axios
Screenshot 2021-03-08 at 12 20 58

- got
Screenshot 2021-03-08 at 12 23 14

This is my final review, I will leave it as a comment instead of rejecting the pull request in case you want to get another review and merge it anyway but in my opinion the way it is handling this issue goes against the good practice.

mrcnski: There are concrete engineering reasons for making this change, and no good reason against it other than "it's not correct" -- which may be your opinion, but it's not a fact that you've managed to demonstrate.

> If my opinion is not enough, think of the industry standards and leaders and how they handle your example:

These are just libraries that process the given data as-is, without context. I don't see the relevance.

dghelm: In isolation, I can see how this is mostly harmless and allows some backward compatibility for an unfortunate mistake in string encoding from legacy code.

That said, David has made comments that have made me think "SkyDB" is moving toward a model where any registry link pointing to a skylink is "SkyDB." If that's the case, staying true to types in `skynet-js` (as they are seen and handled by the server and browser) seems pretty important.

Maybe we do a `stringParse` field in the customOptions parameter but log deprecate warnings when the parsing triggers? So apps that know they have this data can set it on `client` never write that way again and have X amount of time to fix any problem files still on the network that are needed?

mrcnski: > That said, David has made comments that have made me think "SkyDB" is moving toward a model where any registry link pointing to a skylink is "SkyDB." If that's the case, staying true to types in skynet-js (as they are seen and handled by the server and browser) seems pretty important.

Yeah, I agree with that. I was just confused by all the previous talk about axios etc. which is irrelevant here.

I would rather not be adding additional options and complexity just for this case, only to deprecate them later. I think best would be for devs to simply reimplement `getJSON` using `registry.GetEntry` followed by `getFileContents(skylink)` if they are expecting a string result.

PR #353, opened 2021-03-02
Change default getEntry timeout to 2s

mrcnski:
This is a perflormance fix that will make the perceived performance of many applications snappier by default.

The vast majority of registry reads are under 1 s:
https://discord.com/channels/542938080349519882/799542652143403019/803837030903054347
so if a read takes longer than 2 s it is safe to say it will fail.

Specific applications can set the timeout even lower, 1.5 s should be a safe value.


interstella-5555 (dismissed):

dghelm (approved):
LGTM

mrcnski:
Closing since this parameter will be deprecated soon anyway:

> Chris and I are working on an MR to have siad default to better timeouts, once that's in the SDK shouldn't be specifying a timeout at all
siad will pick a timeout based on how long it usually takes to time out

PR #363, opened 2021-03-09
Allow empty registry entry data

mrcnski:


src/registry.ts
```diff
@@ -139,10 +139,16 @@ export async function getEntry(
     );
   }
 
+  let data;
```
MSevey: nit: can't you do
```
let data = "";
if (response.data.data) {
  data = Buffer.from(hexToUint8Array(response.data.data)).toString();
}
```

MSevey (dismissed):

interstella-5555 (approved):

MSevey (approved):

Issue #183, opened 2020-11-11
Reorganize utils


mrcnski:
The `utils.ts` file is stuffed with a bunch of different functions by this point. I think it would be itneresting to organize them into a `utils` directory with different files (for example, theres a few functions for manipulating URLs that can go into `utils/url.ts`).

emersonmacro:
Hello, I'm new to Skynet but I'm interested in getting more involved. Made a PR for this issue - feedback welcome.

mrcnski:
Closed in #364

PR #364, opened 2021-03-15
utils: reorg

emersonmacro:
Took a stab at issue https://github.com/NebulousLabs/skynet-js/issues/183


mrcnski:
Thank you for your contribution! There may be a delay in getting this reviewed, as we have a very busy period right now. We'll get to it as soon as possible. The overall structure of the PR is a 👍 from me

src/utils/skylink.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// TODO: This will be smarter. See
+// https://github.com/NebulousLabs/skynet-docs/issues/21.
+/**
+ * Returns the default portal URL.
+ *
+ * @returns - The portal URL.
+ */
+export function defaultPortalUrl(): string {
```
MSevey: I would put this in the url file.

src/utils/utils.test.ts
```diff
@@ -1,22 +1,16 @@
+import { getFileMimeType, getRelativeFilePath, getRootDirectory } from "./file";
```
MSevey: We should also split this test file out so we have `url.ts` and `url.test.ts` for example.

MSevey (changes requested):

src/utils/skylink.ts
emersonmacro: Good call. I wasn't sure where that one should go. PR updated.

src/utils/utils.test.ts
emersonmacro: Split out tests for each file 👍

emersonmacro:
Not sure why some of the CI builds failed but the 15.x build was successful... `yarn test` passed all tests for me locally, running node 12.x. Any suggestions?

MSevey (approved):

MSevey:
> Not sure why some of the CI builds failed but the 15.x build was successful... `yarn test` passed all tests for me locally, running node 12.x. Any suggestions?

Going to have to look to Marcin on that one.

src/utils/skylink.ts
```diff
@@ -0,0 +1,162 @@
...
+import parse from "url-parse";
+import { CustomClientOptions } from "../client";
+import { trimForwardSlash, trimSuffix, trimUriPrefix } from "./string";
+
+/**
+ * Base custom options for methods hitting the API.
+ *
+ * @property [endpointPath] - The relative URL path of the portal endpoint to contact.
+ */
+export type BaseCustomOptions = CustomClientOptions & {
```
mrcnski, line 12: The options stuff like `BaseCustomOptions` and `defaultOptions`. doesn't seem to fit in `skylink.ts`. Maybe `utils/options.ts`? `ParseSkylinkOptions` should stay here.

mrcnski (approved):
I had one small comment, but it's not blocking. The PR looks good as is. Thank you!

PR #377, opened 2021-03-18
Add withCredentials: true to all axios calls

### LINKED ISSUE #375: Include credentials with Axios call
GET and POST calls need cookies sent with Accounts.

`fetch("https://siasky.xyz/AACU3d_HIQ8jkSPdQwn415O0AVvvP6jo9RALPXWUltwcfQ", { credentials: 'include'})`

For axios: 
[https://stackoverflow.com/a/43178070](https://stackoverflow.com/a/43178070)

mrcnski:
Closes #375

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -163,6 +163,8 @@ export class SkynetClient {
 
       maxContentLength: Infinity,
       maxBodyLength: Infinity,
+      // Allow cross-site cookies.
+      withCredentials: true,
     });
   }
 }
--- a/src/upload.test.ts
+++ b/src/upload.test.ts
@@ -37,6 +37,14 @@ describe("uploadFile", () => {
     expect(data.skylink).toEqual(sialink);
   });
 
+  it("should set 'credentials' to 'include'", async () => {
+    await client.uploadFile(file);
+
+    expect(mock.history.post.length).toBe(1);
+    const request = mock.history.post[0];
+    expect(request.withCredentials).toBeTruthy();
+  });
+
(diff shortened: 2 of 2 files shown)


dghelm:
To make sure this works correctly, we need to follow this idea --

We include credentials by default.
If someone hard-coded to be siasky.net, it still works if on siasky.net
If someone is on localhost and hardcodes a portal, credentials won't be included.
if someone is on skyportal.xyz and requests at siasky.net, it won't included.

mrcnski:
Well I think the portal argument should be removed in favor of a `devPortal` option, which only takes effect if you're not running from a portal already

Ideally the case of portals not matching should never happen.

For now I think we just need to check if the current URL matches the portal we are making requests to (or is a subdomain of it), if not don't set the credentials

dghelm:
I think the error is coming from siasky.net server settings. Once corrected, calling out to outside domains shouldn't cause an error, it just won't include the cookie. I think the commit is good "as-is", but we may need to revisit the inline comment. I think this is "same-site, cross-origin" where the cookies get communicated.

dghelm (approved):

interstella-5555 (approved):

dghelm:
Karol expressed concerns that this may throw errors on portals that haven't updated:
https://github.com/NebulousLabs/skynet-webportal/blob/master/docker/nginx/conf.d/include/cors#L2-L3
https://github.com/NebulousLabs/skynet-webportal/blob/master/docker/nginx/conf.d/include/cors#L12-L13

Let's keep a close eye on this once it's in beta.

PR #383, opened 2021-03-24
Update version to 3.0.2

mrcnski:
I want to release a new version with an addition which is needed for `skybin`. Also, it's been a while since we released a new stable version and there have been some useful fixes.

**Note:** We skip version 3.0.1 because we never published a corresponding stable version for 3.0.1-beta and there has been a new beta release since then.


dghelm:
No issues here.

PR #385, opened 2021-03-27
[BREAKING] Add async portal url init, get*UrlForPortal functions

mrcnski:
- All client download and `client.get*Url` functions are now `async`.
- Added `getEntryUrlForPortal` standalone method
- Added `getSkylinkUrlForPortal` standalone method

**Note:** To keep the PR lean I didn't address the case of providing a portal URL (e.g. `new SkynetClient("https://siasky.net")`) when we are already on a portal -- that should have no effect, the portal argument should only be for dev testing.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -62,8 +62,8 @@ export type RequestConfig = CustomClientOptions & {
  * The Skynet Client which can be used to access Skynet.
  */
 export class SkynetClient {
-  // TODO: This is currently the url of the skapp and not the portal. It should be the value of 'skynet-portal-api' header. This will be a promise, which will be a breaking change.
-  portalUrl: string;
+  // The value of the 'skynet-portal-api' header.
+  portalUrl: Promise<string>;
   customOptions: CustomClientOptions;
 
   // Set methods (defined in other files).
@@ -105,11 +105,29 @@ export class SkynetClient {
    * The Skynet Client which can be used to access Skynet.
    *
    * @class
-   * @param [portalUrl] The portal URL to use to access Skynet, if specified. To use the default portal while passing custom options, use ""
+   * @param [initialPortalUrl] The initial portal URL to use to access Skynet, if specified. A request will be made to this URL to get the actual portal URL. To use the default portal while passing custom options, pass "".
    * @param [customOptions] Configuration for the client.
    */
-  constructor(portalUrl: string = defaultPortalUrl(), customOptions: CustomClientOptions = {}) {
-    this.portalUrl = portalUrl;
+  constructor(initialPortalUrl = "", customOptions: CustomClientOptions = {}) {
+    // Get the initial portal URL, if not provided.
+    if (initialPortalUrl == "") {
+      initialPortalUrl = defaultPortalUrl();
+    }
+    // Kick off async request for the portal URL.
+    this.portalUrl = this.executeRequest({
+      ...customOptions,
+      method: "head",
+      url: initialPortalUrl,
+      endpointPath: "/",
+    }).then((response) => {
+      if (typeof response.headers === "undefined") {
+        throw new Error(
+          "Did not get 'headers' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+        );
+      }
+      return response.headers["skynet-portal-api"] ?? "";
+    });
+
     this.customOptions = customOptions;
   }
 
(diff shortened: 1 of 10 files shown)


src/utils/url.ts
```diff
@@ -56,3 +61,141 @@ export function addUrlQuery(url: string, query: Record<string, unknown>): string
...
+ * @throws - Will throw if the provided timeout is invalid or the given key is not valid.
+ */
+export function getEntryUrlForPortal(
+  portalUrl: string,
+  publicKey: string,
+  dataKey: string,
+  customOptions?: CustomGetEntryOptions
+): string {
+  /* istanbul ignore next */
+  if (typeof portalUrl !== "string") {
```
peterjan, line 82: some helper like this might clean up the code a bit I think
```
function validateParamIsString(name: string, value: string) {
    if (typeof value !== 'string') {
         throw new Error(`Expected parameter '${name}' to be type string, was type ${typeof value}`);
    }
}

export function getEntryUrlForPortal(...) {
    // validate input params
    validateParamIsString(portalUrl)
    validateParamIsString(publicKey)
    validateParamIsString(dataKey)
}
```

src/client.ts
```diff
@@ -62,8 +62,8 @@ export type RequestConfig = CustomClientOptions & {
  * The Skynet Client which can be used to access Skynet.
  */
 export class SkynetClient {
-  // TODO: This is currently the url of the skapp and not the portal. It should be the value of 'skynet-portal-api' header. This will be a promise, which will be a breaking change.
-  portalUrl: string;
+  // The value of the 'skynet-portal-api' header.
+  portalUrl: Promise<string>;
```
peterjan, line 66: this is pretty unconventional but I'll let Karol be the judge

src/client.ts
```diff
@@ -105,11 +105,34 @@ export class SkynetClient {
...
    *
    * @class
-   * @param [portalUrl] The portal URL to use to access Skynet, if specified. To use the default portal while passing custom options, use ""
+   * @param [initialPortalUrl] The initial portal URL to use to access Skynet, if specified. A request will be made to this URL to get the actual portal URL. To use the default portal while passing custom options, pass "".
    * @param [customOptions] Configuration for the client.
    */
-  constructor(portalUrl: string = defaultPortalUrl(), customOptions: CustomClientOptions = {}) {
-    this.portalUrl = portalUrl;
+  constructor(initialPortalUrl = "", customOptions: CustomClientOptions = {}) {
+    // Get the initial portal URL, if not provided.
```
peterjan, line 112: what scenario prevents us from defaulting to `defaultPortalUrl` here in the params?

src/client.ts
```diff
@@ -105,11 +105,34 @@ export class SkynetClient {
...
+    // Kick off async request for the portal URL.
+    this.portalUrl = this.executeRequest({
+      ...customOptions,
+      method: "head",
+      url: initialPortalUrl,
+      endpointPath: "/",
+    }).then((response) => {
+      /* istanbul ignore next */
+      if (typeof response.headers === "undefined") {
+        throw new Error(
```
peterjan, line 125: these are all uncaught exceptions I think

src/client.ts
```diff
@@ -105,11 +105,34 @@ export class SkynetClient {
...
+          "Did not get 'headers' in response despite a successful request. Please try again and report this issue to the devs if it persists."
+        );
+      }
+      const portalUrl = response.headers["skynet-portal-api"];
+      if (!portalUrl) {
+        throw new Error("Could not get portal URL for the given portal");
```
peterjan, line 131: these are all uncaught exceptions I think

src/utils/url.ts
peterjan: helper would also get rid of the `/* istanbul ignore next */` everywhere, which'd be nice

mrcnski:
Closing in flavor of https://github.com/SkynetHQ/skynet-js/pull/18. Comments will be addressed there.

src/utils/url.ts
mrcnski: Follow-up: https://github.com/SkynetHQ/skynet-js/issues/19

src/client.ts
mrcnski: Throwing in `.then` is a rejected promise.

peterjan: what do you mean, my point is that if this error gets thrown, it will register as 
`Uncaught (in promise)` in the console log of whatever code or script that's calling this constructor, right?

mrcnski: You may be right, idk. I think `reject` should be used in this case.

